PR #43, opened 2026-06-01
ci(pkgdown): release/dev site split so root reflects the CRAN version

ehrlinger:
## Problem

After the post-acceptance bump to `1.0.3.9000`, the pkgdown site root shows `1.0.3.9000` instead of the CRAN `1.0.3`. The old workflow deployed `main`'s DESCRIPTION version and wiped the whole site each build (`peaceiris` + `publish_dir: ./docs`), with no released-vs-dev split.

## Fix (standard r-lib dev-mode setup)

- **`_pkgdown.yml`**: `development: mode: auto`. Released versions build to the site **root**; development versions (`Version` ending `.9000`) build to **`/dev/`** with a "(development version)" marker. Verified locally — `as_pkgdown()` routes `1.0.3.9000` → `docs/dev/`, a `1.0.3` release → `docs/` root.
- **`pkgdown.yaml`**: build with `build_site_github_pages()` and deploy with `clean: false` (JamesIves action), so a dev build into `/dev/` no longer wipes the released root, and a release build doesn't wipe `/dev/`. Added a `release: published` trigger; deploy guarded to `main` pushes, releases, and `workflow_dispatch`.

## ⚠️ One-time follow-up after merge

Because `main`'s version is `1.0.3.9000` (dev), the workflow only ever builds `/dev/` from `main` — it won't refresh the **root** until a release-version build happens (v1.1.0). So once this is on `main`, the root needs a one-time seed with the `1.0.3` release docs:

```r
# local, on a clean main checkout with this config:
desc::desc_set_version("1.0.3")          # temporarily, do NOT commit
pkgdown::deploy_to_branch(new_process = FALSE)   # release -> gh-pages root
# then discard the DESCRIPTION change
```

After that, the root self-maintains: each future release build refreshes it, and `main`/dev pushes keep `/dev/` current. I can run this step once the PR is merged (it's a live gh-pages deploy), or you can.

This goes to `main` because the live site deploys from `main` (same as the post-acceptance badge swap). It'll reach `dev` at the next sync.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -5,6 +5,8 @@ on:
     branches: [main, master]
   pull_request:
     branches: [main, master]
+  release:
+    types: [published]
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


ehrlinger:
Closing — on reflection we want the pkgdown site to simply track main's current version (the dev HEAD), which is standard and what the existing setup already does. The release/dev split was more machinery than needed. Site continues to reflect main.

PR #48, opened 2026-06-02
test: weighted multiphase + covariates coverage (7c)

ehrlinger:
**Base: `dev`** (v1.1.0 integration branch).

Closes roadmap **7c** "weighted multiphase + covariates": every weighted multiphase
test was previously intercept-only, so the weighted analytic beta-score and the
weighted CoE×covariate interaction were unverified.

Adds four integer-weight duplication-parity tests with a covariate entering **both**
phases (`hzr_phase(formula = ~ x)`):

| Level | Check | File |
|---|---|---|
| Likelihood | `.hzr_logl_multiphase` weighted == duplicated-row | `test-weights.R` |
| Gradient | `.hzr_gradient_multiphase` (weighted) == `numDeriv::grad` | `test-weights.R` |
| Fit, CoE off | `hazard(..., weights, conserve=FALSE)` coef+objective == duplicated | `test-weights.R` |
| Fit, CoE on | same with `conserve=TRUE` (weighted CoE × covariate) | `test-conservation-of-events.R` |

**Test + docs only — no `R/` changes.** All four passed on first run (the implemented
path was correct). Full suite: 0 failures, 1299 passing, 2 pre-existing skips.

Spec + plan: `inst/dev/WEIGHTED-MULTIPHASE-COVARIATES-DESIGN.md`,
`inst/dev/PLAN-weighted-multiphase-covariates.md`.

> Note: an earlier revision of this PR mistakenly targeted `main`, which made the
> diff span the entire v1.1.0 dev cycle. Retargeted to `dev`; the R/source, version,
> NEWS, and vignette items Copilot flagged belong to dev baseline and are handled in
> a separate follow-up PR.

### DIFF
--- a/inst/dev/DEVELOPMENT-PLAN.md
+++ b/inst/dev/DEVELOPMENT-PLAN.md
@@ -527,7 +527,7 @@ or known numerical gaps. Priority order for investigation:
 | **Interval censoring under multiphase** | Code path exists; no real-data or SAS parity fixture | No parity test |
 | **Hessian stability at 12+ parameters** | Numerical Hessian inversion can become ill-conditioned; `hm.death.AVC.deciles` (13 params) passes but borderline | Passing, fragile |
 | **`hzr_competing_risks()` with weights** | Not a hardening item: `hzr_competing_risks(time, event)` has no `weights` argument — the Aalen-Johansen/Greenwood recursion counts raw row tallies. Weighting is an unbuilt **feature** (weighted at-risk + event counts + weighted-Greenwood variance), not a latent bug. Scoped as feature 8a. | Feature → 8a |
-| **Weighted multiphase + covariates** | All weighted parity tests are intercept-only; covariate + weight combination untested | No test |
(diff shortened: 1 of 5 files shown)


ehrlinger:
Thanks @Copilot — addressed:

1. **"Not test-only / version bump / R changes"** — correct symptom, wrong cause: this PR had mistakenly targeted `main`, so the diff spanned the entire v1.1.0 `dev` cycle. **Retargeted to `dev`** — the diff is now 5 files (2 test + 3 docs), genuinely test+docs only. No `R/` or version changes here.
2. **NEWS header `1.1.0` vs `1.1.0.9000`** — dev-baseline, not in this PR's diff. Fixed in #50.
3. **`time_lower` vignette callout stale** — agreed, stale after the PR #36 dual-use fix; dev-baseline. Rewritten in #50.
4. **`hzr_bootstrap()` prefer stored weights over `parent.frame()` eval** — agreed and confirmed the premise (`object$data$weights` is stored at `hazard_api.R:496`); dev-baseline. Fixed in #50 with an out-of-scope regression test.

Items 2–4 were only visible because of the wrong base; all three are handled in #50 against `dev`.

PR #51, opened 2026-06-02
fix(bootstrap): resolve data from stored frame, not caller-frame eval

ehrlinger:
Sibling of the weights fix in #50. `hzr_bootstrap()` reconstructed the original
data with `eval(cl$data, parent.frame())`, which raises "object not found" when
the original `data` symbol is no longer in scope (e.g. the fit was built inside a
helper that has since returned) — bootstrap then failed before the resample loop.

### Why not just use `object$data`
The refit re-runs the **original call** (formula / `Surv` parsing) against a
resampled data.frame, so it needs the full frame's named columns. `object$data`
only carried `time`/`status`/`x`/`weights`, so it can't drive the refit.

### Fix
- `hazard()` now stores the evaluated model frame on **`object$data$frame`**
  (formula path; `NULL` for raw-vector calls).
- `hzr_bootstrap()` resamples that stored frame; falls back to
  `eval(cl$data, parent.frame())` only for objects fitted before the frame was
  stored (backward compat).
- Regression test builds a weighted fit entirely inside `local({...})` so **both**
  the `data` and `weights` symbols are out of scope, then asserts
  `n_success > 0`, `n_failed == 0`. Verified red before the fix
  (`eval(cl$data, parent.frame())`: object 'bdf' not found).

**Verification:** full suite 0 failures, 1303 passing, 2 pre-existing skips; lintr clean.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,7 +19,8 @@
   was built inside a helper that has returned).  Caller-frame evaluation
   remains a fallback for objects fitted before weights were stored.  The same
   fragility applied to the call's `data` argument: `hazard()` now stores the
-  evaluated model frame on the fitted object (`object$data$frame`), and
+  evaluated `data` argument (the data frame passed to `hazard()`, not a
+  `model.frame()` result) on the fitted object (`object$data$frame`), and
   `hzr_bootstrap()` resamples that stored frame instead of re-evaluating
   `cl$data` in `parent.frame()`, so bootstrap succeeds even when the original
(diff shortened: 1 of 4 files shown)


ehrlinger:
Thanks @Copilot — all three comments are the same valid point and are addressed in e3cc9ba. Reworded the two code comments and the NEWS entry: `object$data$frame` is the **evaluated `data` argument** (the data frame passed to `hazard()`, already forced), explicitly **not** a `model.frame()` result — so no implied NA-handling/term-selection semantics. Comments/NEWS only; no behavior change.

PR #52, opened 2026-06-02
feat(predict): per-phase CLs for decompose=TRUE, se.fit=TRUE (7c)

ehrlinger:
Unblocks predict(type='cumulative_hazard', decompose=TRUE, se.fit=TRUE) for multiphase models. Returns long per-phase + total delta-method CLs. Reuses the existing per-phase Jacobian blocks + delta-method sandwich (shared via new .hzr_free_vcov() helper). Spec + plan in inst/dev/.


ehrlinger:
Thanks @Copilot — all three addressed in the latest commit:

1. **Ordered factor** — `component` is now `factor(..., ordered = TRUE)`, matching the Rd / design text. Added `expect_true(is.ordered(...))`.
2. **Guard scope** — the `decompose + se.fit` type check now fires only for multiphase fits (`object$spec$dist == "multiphase"`). Per the `@param decompose` contract, `decompose` is ignored — not errored — for single-distribution models. Added a regression test: a Weibull fit with `decompose = TRUE, se.fit = TRUE, type = "survival"` now returns the standard aggregate frame instead of erroring.
3. **Stale comment** — corrected to "decompose=TRUE, se.fit=FALSE".

Full suite: 0 failures, 1335 passing; lintr clean.

PR #65, opened 2026-06-05
Multiphase fit correctness: vcov() extraction + CoE left-truncation

ehrlinger:
Two multiphase fit-correctness fixes from the v1.1 fixture-parity workstream (gap-list P1 #1 and P1 #6). Both verified against the full test suite (**0 failures / 1460 passed**).

## P1 #1 — `vcov()` usable and named for multiphase fits
`vcov.hazard()` returned a scalar `NA` whenever **any** cell was `NA`, and carried no dimnames otherwise. Multiphase fits legitimately have `NA` variance rows (fixed shapes; the Conservation-of-Events-conserved phase `log_mu`), so `vcov()` was effectively unusable for almost every multiphase model, and callers could not distinguish `early.x` from `constant.x`.

Now returns the full matrix with `NA` rows preserved and rows/cols named by coefficient (phase-prefixed for multiphase). Scalar `NA` only when no covariance matrix exists.

**Note:** the gap-list framed this as a SAS `.lst` "wide-vcov asymmetry / fix cursor-based routing" bug. That was stale — the parser has handled duplicate cross-phase labels correctly since v0.9.8 (`hm.death.patient`, `hm.death.AVC.deciles` both parse symmetric). The real blocker was R-side extraction.

## P1 #6 — Conservation of Events honors left-truncation
For counting-process data (`Surv(start, stop, event)`), CoE conserved `Σ H(stop)` while the likelihood scores intercepts on the entry-time scale `Σ E = Σ [H(stop) − H(start)]`. The conserved phase absorbed the spurious `Σ H(start)`, biasing its intercept and lowering the log-likelihood — the documented `hz.te123.OMC` fit-1 parity offset.

Verified against the C HAZARD source: `consrv.c` fixes one mu per iteration exactly as R does (the gap-doc's "apply to all phases" prescription was a misread of `setcoe_calc_scaling.c`, the one-time initial scaling). The real divergence is the omitted `CF(ST)` term, which the C `setcoe` derivation requires under `LCENSOR`/`STARTTME`.

Fix threads `time_lower` into `.hzr_conserve_events()` / `.hzr_select_fixmu_phase()`, subtracting the per-phase entry-time cumulative hazard. Plain right-censored fits are unaffected. Adds a synthetic left-truncated invariant test.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,17 @@
 
 ## Bug fixes
 
+* **Conservation of Events ignored left-truncation (counting-process entry
+  times).** For multiphase fits on `Surv(start, stop, event)` data, the CoE
+  reparameterization conserved `Sum H(stop)` while the likelihood scores the
+  intercepts on the entry-time scale, `Sum E = Sum [H(stop) - H(start)]`. The
+  conserved phase therefore absorbed the spurious `Sum H(start)`, biasing its
+  intercept and lowering the attained log-likelihood (the `hz.te123.OMC` fit-1
+  parity offset, gap-list P1 #6). `.hzr_conserve_events()` and
+  `.hzr_select_fixmu_phase()` now subtract the per-phase entry-time cumulative
+  hazard, matching the likelihood and C HAZARD `setcoe` under `LCENSOR`/
+  `STARTTME`. Plain right-censored fits (no `start` time) are unaffected.
+
 * **`vcov()` was unusable for multiphase fits and returned an unnamed matrix.**
   `vcov.hazard()` collapsed the entire matrix to a scalar `NA` whenever any cell
   was `NA`. Multiphase fits legitimately have `NA` variance rows -- for
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,18 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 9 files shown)


ehrlinger:
Addressed all four Copilot review items in 4d25a20:

1. **Final CoE adjustment missing `time_lower`** (R/likelihood-multiphase.R) — real bug, fixed. The post-optimization adjustment now passes `time_lower`, so a left-truncated fit's returned conserved `log_mu` is solved on the same entry-time scale as the objective. The original test only checked the objective value (computed correctly during optimization), not the returned coefficients, which is why it slipped through.
2. **Multi-start RNG flakiness** (test-coe-left-truncation.R:58, :75) — `.coe_fit()` now calls `set.seed()` before each fit, so the CoE-vs-free objective comparison is deterministic across platforms/R versions.
3. Added a self-consistency assertion (LL recomputed at the returned coefficients == reported objective) that directly catches item 1.
4. **`time_lower` undocumented** (man/dot-hzr_select_fixmu_phase.Rd) — added the `@param time_lower` roxygen and regenerated the Rd so `\arguments{}` matches `\usage{}`.

Full suite: 0 failures, 1480 passed.

PR #66, opened 2026-06-05
docs: fixture-parity gap list + private fixture-repo scaffold

ehrlinger:
Dev-only scoping material (under `inst/dev/`) for the v1.1+ cross-codebase fixture-parity workstream — prep for the John + Rajes meeting.

- **`FIXTURE-GAP-LIST.md`** — Groups A–E gap inventory + v1.1 prioritization. Baseline: 54/54 SAS parity expectations passing against the 8 primary fixtures.
- **`hazard-fixtures-scaffold/`** — README/PHI policy, repo layout, env-var conventions, and a validate-fixtures workflow for the proposed private `hazard-fixtures` sibling repo.

The second commit corrects two entries against the **verified C HAZARD source**:
- **P1 #1** was not a `.lst` parser bug (the cursor routing already handled duplicate cross-phase labels since v0.9.8); the real fix was R-side `vcov()` extraction (PR #65).
- **P1 #6** was not Lagrange-vs-closed-form. The original Q4 misread `setcoe_calc_scaling.c` (one-time initial scaling) as the per-iteration constraint; `consrv.c` fixes one mu exactly as R does. The real divergence was the omitted left-truncation `CF(ST)` term (PR #65).

### DIFF
--- a/inst/dev/FIXTURE-GAP-LIST.md
+++ b/inst/dev/FIXTURE-GAP-LIST.md
@@ -53,17 +53,21 @@ need to author `.sas` jobs + capture `.lst` output, then wire them into both har
 
 | Feature | R test file | SAS HAZARD statement | Notes |
 |---|---|---|---|
-| Interval censoring (`status = 2`, `Surv(t_low, t_high, event)`) | `test-interval-censoring-weibull.R`, `test-interval-censoring-other-dists.R` | `LCENSOR` + `RCENSOR` pair | Need a dataset where actual intervals exist, not just endpoints. Clinical calibration data (echo re-read windows?) would be ideal. |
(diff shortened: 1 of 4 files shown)


ehrlinger:
Addressed all six Copilot items in f2e23c2:

1. **OMC layout comment** — clarified the `.lst` **is** committed (aggregate stats); only *re-running* the job needs raw omc data.
2. **`HAZARD_OMC_RAW` "never export"** — reworded to "local machine only — never set in CI or commit the path" (the shell `export` vs "never export" read as contradictory).
3. **PHI guard scope** — the workflow now also rejects flat-file formats (`.csv`/`.tsv`/`.txt`/`.sas7bndx`), matching the scaffold `.gitignore` and the step's stated purpose. Also excluded `./.git/*` and renamed the step to "Fail if raw data files exist".
4. **Interval-censoring syntax (B1)** — corrected to `ICENSOR c3_var = ctime_var` per the verified Q1 finding (was `LCENSOR + RCENSOR`).
5. **Distribution table (B2)** — reconciled with Q2: SAS HAZARD has no `DIST=` keyword; log-logistic/log-normal are R-only with no SAS counterpart; exponential only exists as a constant-only degenerate case. Removed the "create SAS jobs" actions that contradicted Q2.
6. **"CI will detect mismatches"** — clarified the validate-fixtures CI only checks that `.lst` files **parse**, not cross-version numerical drift (that's the consuming repos' parity tests' job).

PR #67, opened 2026-06-05
feat: align hzr_deciles() to SAS %DECILES + hm.death.AVC.deciles parity (Group A)

ehrlinger:
Completes the second Group A fixture parity test **and** resolves both scope notes from the prior session by leaning on the C/SAS 4.3.0 source.

## hzr_deciles() aligned to SAS %DECILES (behavior change)
Reworked `hzr_deciles()` to reproduce the SAS `deciles.hazard` macro exactly:
- **All** subjects ranked into equal-sized risk groups by predicted survival at the horizon.
- Per-group **expected** = Σ predicted cumulative hazard at each subject's *own follow-up time* (group totals sum to total observed events under conservation of events).
- `time` now only stratifies into risk groups — it no longer excludes censored subjects, and expected/observed totals are horizon-independent.
- Previously: excluded censored-before-horizon subjects, expected = `Σ(1 − S(horizon))` — a different statistic.
- Output columns unchanged; definitions updated in `?hzr_deciles` and NEWS.

## True decile GOF parity (scope note 2 resolved)
The SAS decile table **was** in the `.lst` all along — under the spaced title `D E C I L E   A N A L Y S I S`, which an earlier `grep "decile"` missed (the "not captured" note was wrong). Added `.hzr_parse_sas_deciles()` and the test now asserts full **CASES / EXPECTED / ACTUAL** parity, reproduced to print precision (62 each; 36.3/18.3/10.7/3.8/0.8; 36/20/11/2/1; total 70/70), plus the CoE identity (total expected = total observed = 70).

## Deterministic, non-circular start (scope note 1 resolved)
The model-parity test now starts from the SAS job's documented `PARMS` values with `n_starts = 1` — **deterministic** (no RNG) and **non-circular** (start where SAS started, converge to where SAS finished), instead of warm-starting from the converged answer. The honest result: R agrees with SAS MLEs to ~1e-4 on a flat likelihood (the prior 1e-4 tolerance was loosened to 1e-3 to reflect this real, tiny R-vs-SAS optimizer difference the circular start was hiding).

The same `.sas`-`PARMS` deterministic-start pattern is the proper fix for the RNG-order-dependent multi-start parity tests (KUL/AVC/OMC) flagged earlier — follow-up task updated.

Full suite: **0 failures, 1504 passed**. Vignette `.html` will pick up the new decile semantics on next render (prose is unchanged and still accurate).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,21 @@
   only that phase's parameters, so per-phase limits do not sum to the total.
   Previously this combination raised an error.
 
+## Changes
+
+* **`hzr_deciles()` now matches the SAS `deciles.hazard` macro exactly.**
+  Previously it excluded subjects censored before the horizon and defined the
+  expected count as `sum(1 - S(horizon))`. It now follows the SAS method: **all**
+  subjects are ranked into equal-sized risk groups by predicted survival at the
+  horizon, and the expected count per group is the **sum of predicted cumulative
+  hazard at each subject's own follow-up time** (so group totals sum to the total
+  observed events under conservation of events). The `time` argument now only
+  stratifies subjects into risk groups; it no longer restricts or excludes any
+  subject, and the expected/observed totals are horizon-independent. Verified to
+  reproduce the `hm.death.AVC.deciles` SAS decile table (CASES/EXPECTED/ACTUAL)
+  to print precision. The output columns are unchanged; their definitions are
+  updated in `?hzr_deciles`.
+
 ## Bug fixes
 
 * **Conservation of Events ignored left-truncation (counting-process entry
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -11,16 +11,20 @@ NULL
 #' compare observed vs. expected event counts in each group.  Good
 #' calibration means the two track each other across the risk spectrum.
 #'
-#' This implements the workflow of the SAS `deciles.hazard.sas` macro.
-#' Patients are ranked by predicted cumulative hazard at a specified time
-#' point, grouped into quantile bins, and each bin is tested with a
-#' chi-square goodness-of-fit statistic. Subjects censored before the
-#' requested horizon are excluded from the observed-vs-expected comparison.
+#' This reproduces the SAS `deciles.hazard.sas` macro. **All** subjects are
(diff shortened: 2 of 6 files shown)


ehrlinger:
Addressed in 2d5689c.

**Copilot item — `par_beta` argument order (test-sas-parity.R:354).** Verified first: the call actually worked, because R matches the named `phase = "Early"` *before* the positional argument, which then fills `name` — so `vapply(names(early_cov), par_beta, numeric(1), phase = "Early")` was correct (the test passed). But it's a real foot-gun, so I reordered `par_beta(name, phase)` (both call sites) to make the positional/named binding unambiguous.

**Folded in the multi-start determinism fix** (the RNG-order task). Rather than a separate PR that would conflict with this one on the same file, I added `set.seed()` to every `test_that` here that fits with `n_starts > 1` (KUL, AVC, VALVES, the three OMC fits). Each test's optimizer basin no longer depends on the RNG state left by the preceding test. Confirmed order-independent: the parity file passes identically after a heavy `rnorm(1e5)` perturbation. With that in place, the `hm.death.AVC.deciles` test no longer needs the save/restore RNG-neutral workaround — simplified to a plain `set.seed()`.

Full suite: 0 failures, 1504 passed.

ehrlinger:
Addressed the three new items in 9f697fc:

1. **`print.hzr_deciles()` stale "excluded" text** (diagnostics.R:151) — under the SAS-aligned method `n_excluded` is always 0, so the print method no longer says "excluded ... censored before horizon"; it now reports that all subjects are included.
2. **Omnibus chi-square `valid`/df** (diagnostics.R:228) — the contributing-group mask is now `expected > 0 & !is.na(chi_sq)`, and `overall_df` is clamped with `max(0L, …)` so degenerate inputs (≤1 contributing group) can't yield a negative df.
3. **Decile-block parse guard** (test-sas-parity.R:426) — the test now asserts the parsed decile table is non-NULL with 6 rows before subsetting, so a missing/unparseable block fails with a clear message instead of "NULL is not subsettable".

(The earlier `par_beta` argument-order comment was already resolved in 2d5689c.)

Full suite: 0 failures, 1506 passed.

PR #68, opened 2026-06-05
test: SAS parity for ac.death.AVC Kaplan-Meier / Nelson-Aalen life tables (Group A)

ehrlinger:
Third Group A fixture — the actuarial life tables (`ac.death.AVC`). Validates `hzr_kaplan()` and `hzr_nelson()` against the SAS `%KAPLAN` / `%NELSONT` macro output.

## Coverage
- **KM overall** (`_CATG=ALL`): n at risk, events, survival, Greenwood SE, and −log(S) cumulative hazard — match to print precision (~5e-6).
- **Nelson-Aalen overall** (`%NELSONT`): cumulative hazard matches.
- **KM stratified by `COM_IV`** (=0, n=154; =1, n=156): n at risk + survival.

## New parser
`.hzr_parse_sas_lifetable()` — header-driven, so it handles both the KAPLAN column layout (`SE_EXACT`) and the NELSONT layout (`NUMRISK`), plus the `_CATG=<n>` strata. Rows are aligned by index since SAS prints event times to 3 decimals.

No model fit — `hzr_kaplan`/`hzr_nelson` are non-parametric and deterministic (no RNG). Skips when fixtures are absent.

Full suite: **0 failures, 1524 passed.**

### DIFF
--- a/tests/testthat/helper-sas-parity.R
+++ b/tests/testthat/helper-sas-parity.R
@@ -450,6 +450,63 @@
   do.call(rbind, rows)
 }
 
+# Parse a SAS %KAPLAN / %NELSONT life-table block. These are printed by the
+# kaplan.sas / nelsont.sas actuarial macros as a wide table with a header line
+#   Obs INT_DEAD NUMBER CENSORED DEAD CUM_SURV SE_EXACT CL_LOWER ... PROPLIFE
+# (KAPLAN) or a slightly different column set (NELSONT adds NUMRISK, drops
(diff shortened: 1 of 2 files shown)


ehrlinger:
Addressed the Copilot item: added `skip_if_not(file.exists(lst), ...)` so the test skips cleanly when an older local fixture set lacks `ac.death.AVC.lst`, instead of hard-erroring in `readLines()`. Parity suite still 0 failures.

PR #69, opened 2026-06-06
test: SAS HAZPRED prediction parity for hp.death.AVC nomogram (Group A)

ehrlinger:
Fourth Group A fixture — `%HAZPRED` prediction parity (`hp.death.AVC`).

## What it does
`%HAZPRED` predicts survival + hazard at a "digital nomogram" of time points from the **saved** `hz.death.AVC` fit (`EXAMPLES.HZDEATH`, the 2-phase Early+Constant model). The test refits that model deterministically from its `.sas` `PARMS` and predicts at the **same exact times** — reconstructed from the SAS `DO` loop (`1*DTY..7*DTY, 14*DTY, 30*DTY, 1,2,3,6,12,18, 24 TO 180 BY 12`, `DTY = 12/365.2425`). Predicting at the *printed* (3-decimal-rounded) months inflates the error ~30×, so exact times matter.

- **`_SURVIV`** via the public `predict(type = "survival")` — matches to ~5e-6.
- **`_HAZARD`** via the internal `.hzr_multiphase_hazard()` — matches to ~1e-5. The instantaneous multiphase hazard is **not a public predict type yet** (`predict(type = "hazard")` is single-dist only), so it's checked through the internal fn.

Adds `.hzr_parse_sas_nomogram()` to read the PROC PRINT prediction table.

## Deferred (noted)
HAZPRED delta-method confidence limits (`_CLLSURV`/`_CLUSURV`/`_CLLHAZ`/`_CLUHAZ`) are **not** asserted — `se.fit` isn't exposed for survival/hazard prediction. Two natural follow-ups surfaced: expose `predict(type = "hazard")` for multiphase, and `se.fit` for survival/hazard (the cumulative-hazard se.fit path from #52 already exists).

Full suite: **0 failures, 1529 passed.** Skips when fixtures absent.

### DIFF
--- a/tests/testthat/helper-sas-parity.R
+++ b/tests/testthat/helper-sas-parity.R
@@ -507,6 +507,32 @@
   df
 }
 
+# Parse the HAZPRED "digital nomogram" prediction table printed by jobs like
+# hp.death.AVC.sas: a PROC PRINT with header
+#   Obs YEARS MONTHS _SURVIV _CLLSURV _CLUSURV _HAZARD _CLLHAZ _CLUHAZ
+# Returns a data frame with those columns (Obs dropped), or NULL if absent.
+.hzr_parse_sas_nomogram <- function(path) {
+  lines <- .hzr_read_lst(path)
+  h <- grep("YEARS[[:space:]]+MONTHS[[:space:]]+_SURVIV", lines)
+  if (!length(h)) return(NULL)
+  cols <- c("YEARS", "MONTHS", "SURVIV", "CLLSURV", "CLUSURV",
+            "HAZARD", "CLLHAZ", "CLUHAZ")
+  rows <- list()
+  for (ln in lines[(h[1] + 1L):length(lines)]) {
+    if (!nzchar(trimws(ln))) next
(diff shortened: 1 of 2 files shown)


ehrlinger:
Both items addressed in 4be27c9 (comment-only — no logic change).

1. **CL rationale was misleading** (test-sas-parity.R:518) — corrected. You're right that `predict(type = "survival", se.fit = TRUE)` is available (multiphase included). I verified the actual reasons the HAZPRED CLs aren't asserted: (a) R's survival CLs don't reproduce SAS HAZPRED's to parity tolerance — a different delta-method transform; even at SAS's 1-SD level (`level = 0.68268948`) the limits differ by ~0.02 — and (b) the *hazard* CLs have no public path since multiphase hazard isn't a `predict()` type. Comment now states this, with both follow-ups (reconcile the survival-CL construction with HAZPRED; expose `predict(type = "hazard")`).
2. **Parser docstring** (helper-sas-parity.R:513) — now notes the returned column names have leading underscores stripped (`SURVIV`, not `_SURVIV`).

Parity suite still 0 failures.

PR #71, opened 2026-06-06
test: Tier-1 correctness invariant suite (R-only, no SAS)

ehrlinger:
Implements **Tier 1** of the correctness strategy (#70, `inst/dev/CORRECTNESS-STRATEGY.md`): single-engine property tests that every fitted model must satisfy by the mathematics — no SAS reference, no captured fixtures. The cheap workhorse of the regime.

Dataset-driven over a model matrix spanning the four single distributions (weibull / exponential / loglogistic / lognormal) **plus** multiphase (conserve on/off, covariates, left-truncation, integer weights).

## Invariants
- **Distributional sanity:** `S ∈ [0,1]` and non-increasing; `H ≥ 0` and non-decreasing; `S = exp(−H)` — which also pins `predict("survival")` and `predict("cumulative_hazard")` to be mutually consistent.
- **Multiphase LL self-consistency:** LL recomputed at the returned coefficients == reported objective — catches the class of the CoE final-adjustment bug from #65.
- **Conservation of Events identity:** `Σ (weighted) events == Σ (weighted) [H(stop) − H(start)]` at the optimum.

All hold across the matrix (CoE exact to ~1e-15). Fast, deterministic, no external dependency — exactly what should run on every CI.

## Deferred (documented in the helper)
Stationarity (score ≈ 0 at the MLE) and analytic-vs-`numDeriv` gradient agreement — the latter is already exercised per-distribution by the analytic-Hessian tests.

Files: `tests/testthat/helper-invariants.R` (generators + assertions), `tests/testthat/test-invariants.R`. Full suite: **0 failures, 1587 passed.**

Depends conceptually on #70 (the strategy); mergeable independently.

### DIFF
--- a/tests/testthat/helper-invariants.R
+++ b/tests/testthat/helper-invariants.R
@@ -1,0 +1,188 @@
+# Tier-1 correctness invariants (see inst/dev/CORRECTNESS-STRATEGY.md).
+#
+# These are R-only, single-engine checks: properties every fitted model must
+# satisfy by the mathematics, with no SAS reference and no captured fixtures.
+# They are the cheap workhorse of the correctness regime -- they caught (or
+# would have caught) the multiphase vcov, CoE left-truncation, deciles-method,
+# and CoE final-adjustment bugs.
+#
+# Covered here: distributional sanity (S in [0,1] and monotone; H monotone and
+# >= 0; S = exp(-H)), predict-type internal consistency, multiphase log-
+# likelihood self-consistency (LL at returned theta == reported objective), and
+# the Conservation-of-Events identity. Future expansion (documented, not yet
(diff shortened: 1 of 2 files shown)


ehrlinger:
Both addressed in the latest commit:

1. **Refitting the matrix 3×** — added a memoized `.inv_models_cached()` (file-scoped env cache); the three invariant loops now use it, so the model matrix is fit once per file run instead of once per `test_that()`. Suite runtime ~0.8s.
2. **CoE tolerance collapsing to 0** — `tolerance = 1e-3 * obs` is now floored at `max(1e-6, 1e-3 * obs)`, so a zero-event group no longer demands exact floating-point equality.

(The 5 inline comments were these 2 distinct issues — 4 of them were the same memoization point on each loop.) Suite still 0 failures.

PR #73, opened 2026-06-06
feat: predict(type='hazard') for multiphase (instantaneous additive hazard)

ehrlinger:
Closes the predict gap noted in #69: multiphase `predict(type = "hazard")` now returns the instantaneous additive hazard `h(t|x) = Σ_j μ_j(x) φ_j'(t)`.

Previously only single-distribution models supported `"hazard"` (`exp(eta)`); multiphase rejected it, and the additive hazard was reachable only through the internal `.hzr_multiphase_hazard()`.

## Behavior
- Time-based like `"survival"`/`"cumulative_hazard"` — requires `newdata$time`; supports covariate `newdata`.
- `se.fit = TRUE` returns delta-method limits on the log scale, via a **numeric** Jacobian of the hazard evaluator (the analytic multiphase Jacobian is the cumulative-hazard one).
- `decompose = TRUE` is rejected for `"hazard"` (the evaluator returns the total only).
- `linear_predictor` remains unsupported for multiphase.

## Validation
- New unit tests (`test-predict-hazard-multiphase.R`): point == internal evaluator, time/covariate handling, `se.fit` limits, decompose/linear_predictor rejection.
- The `hp.death.AVC` SAS parity test now checks `_HAZARD` through the **public** `predict()` path (was the internal fn) and still reproduces the HAZPRED nomogram to ~1e-5.

NEWS + `?predict.hazard` updated. Full suite: **0 failures, 1607 passed**.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,16 @@
 
 ## New features
 
+* `predict.hazard(type = "hazard")` now works for **multiphase** models,
+  returning the instantaneous additive hazard
+  `h(t|x) = sum_j mu_j(x) phi_j'(t)` (previously only single-distribution
+  models supported `"hazard"`, via `exp(eta)`). Like `"survival"` /
+  `"cumulative_hazard"` it is time-based (requires `newdata$time`), supports
+  covariate `newdata`, and `se.fit = TRUE` (delta-method limits on the log
+  scale via a numeric Jacobian of the hazard evaluator). `decompose = TRUE` is
+  not supported for `"hazard"`. This gives the multiphase instantaneous hazard a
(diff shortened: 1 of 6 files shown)


ehrlinger:
Addressed: updated the hp.death.AVC header comment. This PR adds the public multiphase `predict(type = "hazard")` + `se.fit`, so the 'hazard CLs have no public path' note was stale. The comment now states accurately that both survival and hazard have public `predict()`/`se.fit`, and the CLs aren't asserted only because R's transform differs from SAS HAZPRED's (complementary-log-log vs logit) — with a `conf.type = "logit"` option noted as the follow-up. Comment-only; parity suite still 0 failures.

PR #74, opened 2026-06-06
fix: full-information vcov for Conservation-of-Events fits

ehrlinger:
Resolves the conserved-μ variance item (correctness strategy §3). CoE removes one phase's `log_mu` from the optimizer **search** — its score equation *is* the CoE constraint — but the code also dropped it from the reported **uncertainty**: the conserved `log_mu` got an `NA` standard error, and anything depending on it (other SEs, `se(H)`, prediction confidence limits) was understated wherever that phase contributed.

## Fix
At the optimum the CoE solution **is** the unconstrained MLE (the conserved-μ score=0 condition is exactly the CoE constraint). So the reported uncertainty should be the full-information vcov. After convergence, recompute the vcov from the **unconstrained-objective** Hessian (`logl_fn_pre_coe`) over the full free set with the conserved `log_mu` restored — exactly what an all-μ-free (`conserve = FALSE`) fit yields at the same point. `fixed_mask` is updated so the conserved `log_mu` is fixed only for the *search*, not for *inference*.

## Validation
On `hz.death.AVC`, **every** parameter SE now matches the SAS HAZARD reference:
- conserved early `log_mu` (E0): **0.133** vs SAS 0.1330 (was ~0.059)
- constant `log_mu` (C0): 0.4561 vs 0.4561; early `log_t_half` (E2): 0.3605 vs 0.3605

The parity test's previously-noted E0 gap is resolved and now asserted (plus C0). This also fixes `se(H)` everywhere, so combined with the logit transform the HAZPRED survival CLs reproduce SAS to ~8e-6 (the logit CL **option** follows in a separate PR).

Full suite: **0 failures, 1594 passed** — no test relied on the old NA behavior.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,6 +47,9 @@
   conserved `log_mu`), matching an all-`mu`-free (`conserve = FALSE`) fit at the
   same point. On `hz.death.AVC` every parameter SE now matches the SAS HAZARD
   reference (e.g. the conserved early `log_mu`: 0.133 vs the previous ~0.059).
+  The recomputation uses `numDeriv` (Suggests) and an invertible Hessian; if
+  either is unavailable the fit emits a warning and the conserved `log_mu`
+  retains an `NA` standard error (as before).
 
 * **Conservation of Events ignored left-truncation (counting-process entry
   times).** For multiphase fits on `Surv(start, stop, event)` data, the CoE
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,18 @@
 
 ## Bug fixes
 
+* **Conservation-of-Events fits now report the full-information variance.**
+  CoE removes one phase's `log_mu` from the optimizer *search* (its score
+  equation is the CoE constraint), but the previous code also dropped it from
+  the *uncertainty* -- the conserved phase got an `NA` standard error, and
+  anything depending on it (other SEs, `se(H)`, prediction confidence limits)
+  was understated wherever that phase contributed. At the optimum the CoE
+  solution is the unconstrained MLE, so `vcov()` is now recomputed from the
+  unconstrained-objective Hessian over the full free set (including the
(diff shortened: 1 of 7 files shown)


ehrlinger:
Addressed: corrected the stale 'CoE-conserved `log_mu` has NA variance/SE' claims that this PR invalidates.
- `vcov.hazard` `@return` (+ regenerated Rd) and its inline comment: fixed params (e.g. fixed shapes) are NA, but the CoE-conserved `log_mu` now carries a variance (removed from the *search*, full-information vcov at the optimum).
- `.hzr_free_vcov()` header: notes the conserved `log_mu` is not NA and participates in the sandwich.

Doc-only; no behavior change beyond this PR's vcov fix.

ehrlinger:
Addressed all five items — they're one real concern: the full-information vcov recomputation is conditional (`numDeriv` from Suggests + an invertible Hessian), but it fell back **silently** to the reduced vcov while the docs/NEWS described the full-info behavior unconditionally.

- **Warn on fallback** (R/likelihood-multiphase.R): if the recomputation can't run (`numDeriv` missing or Hessian not invertible), the fit now emits a clear warning that the conserved `log_mu` SE stays `NA` and downstream SEs/CLs may be understated. The normal path (numDeriv present, Hessian fine) does **not** warn — verified.
- **Qualified the docs/NEWS** so they match the actual behavior: `vcov.hazard` `@return` + inline comment, the `.hzr_free_vcov()` header, and the NEWS entry now say the conserved `log_mu` carries a variance *when computable* (requires numDeriv + invertible Hessian) and otherwise stays `NA` with a warning.

I kept `numDeriv` in Suggests rather than promoting it to Imports — this is consistent with the existing behavior (the optimizer's post-fit Hessian already degrades to `NA` vcov without numDeriv), and the warning makes the degradation visible.

Full suite: 0 failures, 1609 passed.

PR #117, opened 2026-08-17
fix: three blind spots in stepwise/bootstrap variable selection

### LINKED ISSUE #114: hzr_bootstrap(): selection silently returns nothing when the base fit's stored formula is a symbol
## Summary

`hzr_bootstrap()` in selection mode silently selects **nothing** when the base fit was
created with the model formula held in a variable rather than written literally at the
`hazard()` call. It reports `n_success = n_boot`, `n_failed = 0`, no error and no warning.
The two models are otherwise identical.

## Reproduction

Self-contained, simulated data.

```r
library(TemporalHazard); library(survival)
set.seed(1)
n  <- 600
df <- data.frame(x1 = rnorm(n), x2 = rnorm(n), x3 = rnorm(n))
df$t <- rexp(n, rate = 0.15 * exp(0.9 * df$x1))
df$d <- as.integer(df$t < 10); df$t <- pmin(df$t, 10)

ph <- list(early = hzr_phase("cdf", t_half = 0.5, nu = 1, m = 1,
                             fixed = c("t_half", "nu", "m")),
           late  = hzr_phase("g3", tau = 1, gamma = 1, alpha = 1, eta = 1,
                             fixed = c("tau", "gamma", "alpha", "eta")))
sc  <- list(early = ~ x1 + x2 + x3, late = ~ x1 + x2 + x3)
ctl <- list(n_starts = 1, maxit = 2000)
new_covs <- function(b, base) setdiff(unique(b$replicates$parameter), names(coef(base)))

# (A) formula written literally
a  <- hazard(survival::Surv(t, d) ~ 1, data = df, dist = "multiphase",
             phases = ph, fit = TRUE, control = ctl)
ba <- hzr_bootstrap(a, n_boot = 5, seed = 1, scope = sc, criterion = "score",
                    direction = "forward", slentry = 0.2, slstay = 0.2)

# (B) identical model, formula in a variable
f  <- survival::Surv(t, d) ~ 1
b  <- hazard(f, data = df, dist = "multiphase", phases = ph, fit = TRUE, control = ctl)
bb <- hzr_bootstrap(b, n_boot = 5, seed = 1, scope = sc, criterion = "score",
                    direction = "forward", slentry = 0.2, slstay = 0.2)

new_covs(ba, a)   # 6 covariates
new_covs(bb, b)   # character(0)
```

Output:

```
A literal : stored formula = survival::Surv(t, d) ~ 1 | n_success 5 n_failed 0 | selected: early.x3, early.x1, late.x1, early.x2, late.x2, late.x3
B variable: stored formula = f                        | n_success 5 n_failed 0 | selected: NONE
```

## Cause, as far as I traced it

`hzr_bootstrap()` re-evaluates the base fit's stored call for every replicate, and
`.hzr_refit_with_scope()` rewrites that call's formula to `. ~ . + <candidate>` to test an
entry. When `object$call$formula` is a **symbol** rather than a formula literal, that
rewrite does not survive: the refit errors, `.hzr_stepwise_forward_step_score()` catches it,
reports "nothing accepted", and the forward screen halts at step 1 in every replicate.

The per-replicate warnings are inside `suppressWarnings()`, so the only visible trace is a
single "post-entry refit failed" from the pilot selection.

## Why it is worth guarding

The failure is indistinguishable from success in the returned object. On a real screen this
consumed roughly a day of compute across 25 processes and produced a summary table in which
every row read 100%, which looks like a table of perfectly reliable variables rather than a
table of nothing. Holding a formula in a variable is ordinary R style, and nothing in the
documentation suggests it is unsafe here.

## Suggested fix

Either would have prevented it:

1. **Resolve it.** In `hzr_bootstrap()`, if `cl$formula` is a symbol, replace it with its
   value before the replicate loop: `cl$formula <- eval(cl$formula, envir = object$call_env)`.
   This makes the ordinary style work.
2. **Refuse it.** Check that the stored formula is a call to `~` and stop with a message
   naming the cause, rather than returning an empty screen.

A related detector is filed separately: a select-mode run in which no replicate selects any
covariate currently warns about nothing.

## Session

R 4.6.0, TemporalHazard 1.2.0.

### LINKED ISSUE #115: hzr_bootstrap(): a select-mode run that selects zero covariates reports no warning
## Summary

In selection mode (`scope` supplied), a run in which **no replicate selects any covariate**
returns an object indistinguishable from a successful screen: `n_success = n_boot`,
`n_failed = 0`, no error, no warning. The summary table then contains only the base model's
parameters, every one at `pct = 100`, which reads as a set of perfectly reliable variables.

## Reproduction

The (B) case in the sibling issue produces it, but the cause does not matter for this
report: any route to an empty screen returns the same shape. A zero-covariate result is
worth flagging whatever produced it, since the legitimate cases (an entry criterion too
strict for the data) are also worth telling the user about.

```r
new_covs <- setdiff(unique(bb$replicates$parameter), names(coef(b)))
length(new_covs)   # 0
bb$n_success       # 5   (the maximum)
bb$n_failed        # 0
```

## Why it is worth guarding

A selection frequency is the whole deliverable of a bootstrap screen, so "nothing was ever
selected" is not a quiet edge case, it is the result being empty. Because the base
parameters appear in every replicate by construction, they occupy the summary at 100% and
the table looks populated. A reader has to know to compare the parameter names against
`names(coef(base))` to notice, and there is nothing in the output prompting them to.

## Suggested fix

When `scope` was supplied and
`length(setdiff(unique(replicates$parameter), names(coef(object)))) == 0`, emit a warning
that names the likely causes: an entry criterion stricter than intended, a scope naming
columns absent from the data, or a base fit whose stored call cannot be rewritten (the
sibling issue).

## Session

R 4.6.0, TemporalHazard 1.2.0.

ehrlinger:
Closes #114. Closes #115.

## #114 — the formula arrives as a symbol

`hazard()` records its call with `match.call()`, so the `formula` argument is stored **unevaluated**. Written literally it is a `call`; assigned to a variable first it is a **symbol**:

```r
a <- hazard(Surv(t, d) ~ 1, data = df, ...)   # class(a$call$formula) == "call"
f <- Surv(t, d) ~ 1
b <- hazard(f, data = df, ...)                # class(b$call$formula) == "name"
```

`.hzr_refit_with_scope()` recovered it with `as.formula(deparse(raw_formula))`. For a symbol, `deparse()` yields the string `"f"`, and:

```
Error in formula.character(object, env = baseenv()) : invalid formula "f": not a call
```

Every post-entry refit threw, so no candidate ever entered. The run reported `n_success = n_boot`, `n_failed = 0`, no error, no warning — and a summary holding only the base model's parameters.

**Fix:** evaluate the stored formula in the fit's recorded `call_env`, which handles both forms. A stored formula that fails to resolve now errors naming the problem rather than degrading to an empty screen.

`hzr_stepwise()` was affected directly by the same code path, not only through `hzr_bootstrap()`.

## #115 — an empty screen looked like a successful one

A selection frequency is the entire deliverable of a select-mode run, so "nothing was selected" is the result being empty. It cannot be read off the object: the base model's parameters appear in every replicate by construction, so they occupy the summary at `pct = 100` and the table reads as a set of perfectly reliable variables.

`hzr_bootstrap()` now warns when a select-mode run selects no covariate, naming the likely causes. Legitimate empty screens warn too — an entry criterion no candidate can clear is also worth reporting.

## Verification

Both fixes were written test-first and watched fail.

- #114 RED: `Error ... invalid formula "f": not a call`, and the screen comparison showed the literal run selecting `early.age` while the symbol run selected nothing.
- #115 RED: `Expected ... to throw a warning.` A companion test asserts a *populated* screen stays warning-free, so the guard cannot fire indiscriminately.
- The issue's own reproduction now gives **6 new covariates for both** interfaces, against 6 vs 0 before.

Full suite: **0 failures, 96 documented `skip_on_cran`.**

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -114,6 +114,30 @@ selects.
   replicate is evaluated in a child of it that carries the resampled data and
   weights. Affects both `refit` and `select` modes.
 
+* **`hzr_bootstrap(scope = ...)` selected nothing when the base fit's formula
+  was passed by symbol.** `hazard()` records its call with `match.call()`, so a
+  formula assigned to a variable first (`f <- Surv(t, d) ~ 1; hazard(f, ...)`)
+  is stored as a *symbol* rather than a call. The scope-mutating refit
+  recovered it with `as.formula(deparse(...))`, which turns that symbol into
+  the string `"f"` and errors with `invalid formula "f": not a call`. Every
+  post-entry refit therefore failed, no candidate ever entered, and the run
+  reported `n_success = n_boot`, `n_failed = 0`, no error and no warning --
+  with a summary holding only the base model's parameters. The stored formula
+  is now evaluated in the fit's recorded calling environment, which handles
+  the literal and by-symbol forms alike, and a stored formula that fails to
+  resolve raises an error naming the problem instead of degrading to an empty
+  screen. The same defect affected `hzr_stepwise()` directly. (#114)
+
+* **A select-mode `hzr_bootstrap()` run that selects no covariate now warns.**
+  The base model's own parameters appear in every replicate by construction,
+  so they fill the summary at `pct = 100` and an empty screen reads as a set
+  of perfectly reliable variables; nothing in the output prompted the reader
+  to compare the parameter names against `names(coef(object))`. The warning
+  names the likely causes: an entry criterion stricter than intended, a
+  `scope` naming columns absent from the data, or a base fit whose stored call
+  cannot be rewritten. Legitimate empty screens warn too -- an entry criterion
+  no candidate can clear is also worth reporting. (#115)
+
 * `hzr_bootstrap()` no longer floods the console with per-replicate numerical
   warnings (e.g. ill-conditioned-Hessian notes from unstable resamples), which
   are not individually actionable when the bootstrap aggregates over replicates.
--- a/NEWS.md
+++ b/NEWS.md
@@ -138,6 +138,33 @@ selects.
   cannot be rewritten. Legitimate empty screens warn too -- an entry criterion
   no candidate can clear is also worth reporting. (#115)
 
+* **A stepwise screen that could not score anything now says so, instead of
+  looking like one that finished.** Under `criterion = "score"` a candidate
+  whose Q statistic cannot be computed -- a degenerate or collinear column,
+  or an information matrix that will not invert on this data -- yields `NA`
+  and is dropped from consideration. When that happened to every remaining
+  candidate the step returned exactly what a legitimate "no candidate met
+  `slentry`" stop returns, so a screen that stopped because it was *unable
+  to test* its candidates was indistinguishable from one that tested them
+  and found nothing. The per-step diagnostic existed on the returned object
+  the whole time and had no readers.
+
+  `hzr_stepwise()` now warns when a run stops this way and reports
+  `$criteria$n_uncomputable_scores`. Because `hzr_bootstrap()` runs each
+  replicate under `suppressWarnings()` -- deliberately, so per-replicate
+  numerical noise does not swamp the console -- that warning cannot surface
+  in the mode where it matters most, so the count is aggregated instead:
+  `hzr_bootstrap()` gains `$n_uncomputable_replicates` and warns once when it
+  is non-zero. A replicate that scored nothing still counts toward
+  `n_success` while contributing no selections, so it silently depresses
+  every reported selection frequency -- which is the whole deliverable of a
+  bootstrap screen.
+
+  Found by a pre-release review pass, not by a failing test: the package's
+  own `print.hzr_bootstrap` test runs a five-replicate screen in which four
+  replicates cannot score a candidate and none selects anything, and it
+  passed throughout because it only ever asserted the printed label.
+
 * `hzr_bootstrap()` no longer floods the console with per-replicate numerical
   warnings (e.g. ill-conditioned-Hessian notes from unstable resamples), which
   are not individually actionable when the bootstrap aggregates over replicates.
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -1295,6 +1295,12 @@ print.hzr_nelson <- function(x, digits = 4, ...) {
 #'     and the other statistics are conditional on selection.}
 #'   \item{n_success}{Number of successfully converged replicates.}
 #'   \item{n_failed}{Number of replicates that failed to converge.}
+#'   \item{n_uncomputable_replicates}{Select mode only: number of otherwise
+#'     successful replicates whose screen stopped because no remaining
+#'     candidate's score statistic could be computed, rather than because no
+#'     candidate met `slentry`. Such replicates contribute no selections, so
+#'     a non-zero count means every reported selection frequency is
+#'     depressed. Always `0` in refit mode.}
 #'   \item{mode}{`"refit"` (fixed-formula bootstrap) or `"select"`
 #'     (embedded stepwise selection).}
 #'   \item{scope}{Only present when `mode == "select"`: the candidate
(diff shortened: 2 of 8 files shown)


ehrlinger:
Applied the one finding from a local review pass (Copilot is erroring out during [the GitHub incident](https://www.githubstatus.com/)).

**`R/stepwise-refit.R` — the new not-a-formula guard was unreachable.** `eval()` raises first when the stored symbol has no binding, so the failure surfaced as a bare `object 'f' not found` — naming neither the fit at fault nor the remedy — and `hzr_stepwise()`'s per-candidate `tryCatch()` reduced it to a generic "candidate refit failed" plus an empty screen.

Found by round-tripping a fit through `saveRDS()`/`readRDS()` across two R sessions, which is how this package's production caller loads base fits. Before:

```
ERROR: object 'f' not found
```

After:

```
ERROR: `current`'s stored model formula (f) could not be resolved: object 'f' not
found. The formula was passed to `hazard()` by variable and that binding is no
longer reachable -- for example after saving and reloading the fit in a new
session. Refit the base model with the formula written at the `hazard()` call.
```

This also makes the NEWS entry accurate as written — it already claimed a stored formula that fails to resolve "raises an error naming the problem instead of degrading to an empty screen", which was true only for the resolved-but-wrong-type case until now.

The regression test blanks `call_env` to reproduce the reloaded-fit state in-process, so it needs no second session.

Suite: 0 failures, 1515 passing, 96 documented `skip_on_cran`. `lintr::lint_package()` clean.

ehrlinger:
Third fix added from the bounded pre-release review pass. Same subsystem and same unreleased version, so it rides here rather than conflicting in `R/diagnostics.R`.

## A screen that could not test anything looked like one that finished

Under `criterion = "score"` — **the new default in 1.2.0** — a candidate whose Q statistic cannot be computed (degenerate/collinear column, information matrix that will not invert on this resample) yields `NA` and drops out of `valid`. When that happened to every remaining candidate, the step returned exactly what a legitimate "no candidate met `slentry`" stop returns.

Two opposite meanings, one result:

- *tested them, none was good enough* → the screen worked
- *could not test them* → the screen is broken

`all_scores` held the data to tell them apart and had **zero readers** anywhere in the package.

## Why the obvious fix doesn't work

`hzr_bootstrap()` wraps each replicate in `suppressWarnings()` — deliberately, added in #112 so per-replicate Hessian noise doesn't swamp the console. So a step-level warning is invisible in exactly the mode where this matters. The count is aggregated instead:

- `hzr_stepwise()` — warns, exposes `$criteria$n_uncomputable_scores`
- `hzr_bootstrap()` — gains `$n_uncomputable_replicates`, warns once when non-zero

A replicate that scored nothing still counts toward `n_success` while contributing no selections, so it silently depresses every selection frequency — the whole deliverable of a bootstrap screen.

## It found something immediately

The new warning fired on an **existing test** on first run. `print.hzr_bootstrap reports the mode` runs a five-replicate screen where **four replicates cannot score a candidate and none of the five selects anything**. Verified directly against the resamples:

```
rep 1: uncomputable=TRUE   n=1   selected=<none>
rep 2: uncomputable=TRUE   n=2   selected=<none>
rep 3: uncomputable=FALSE  n=0   selected=<none>
rep 4: uncomputable=TRUE   n=1   selected=<none>
rep 5: uncomputable=TRUE   n=2   selected=<none>
```

It passed throughout because it only ever asserted the printed label. That assertion is now explicit.

On the **full** `avc` data every candidate scores fine (`age` p≈1.8e-4, `mal` p≈2.0e-5) — the failures are specific to resamples. Worth knowing independently of this PR, since score is the new default.

Suite: **0 failures, 1528 passing** (from 1515), 96 documented `skip_on_cran`. `lintr::lint_package()` clean.

Issue #130, opened 2026-08-18
score criterion declines the strongest candidates: observed information goes indefinite at beta = 0


ehrlinger:
Found while building the regression fixture for #129 (the multiphase score-information fallback). **Unrelated to that fix, and pre-existing** — it reproduces on plain right-censored data through the fully-supported analytic path.

## The behaviour

`criterion = "score"` — the 1.2.0 default — scores a candidate with

```
v_beta = I_bb - I_bt %*% solve(I_tt) %*% I_tb        # Schur complement
Q      = u_beta^2 / v_beta
```

evaluated at the current model's MLE with `beta = 0`. That is a Schur complement of the **observed** information. When the candidate's true effect is far from zero, the log-likelihood is *convex* in `beta` at `beta = 0`, the block goes indefinite, `v_beta` turns negative, and `.hzr_score_q()`'s guard correctly returns `NA` rather than a nonsense statistic.

The consequence is that **the criterion declines candidates in proportion to how predictive they are.**

## Measured

2-phase fit (`cdf` early with fixed shapes + `constant`), n = 260, single candidate `z` with true coefficient `beta_true`, entered on the early phase. Analytic and numeric Hessians agree to 1e-4, and interval-censored vs right-censored agree to ~1e-2, so neither is implicated.

| true β | u_β | I_bb | v_β | score Q | wald |
|---|---|---|---|---|---|
| 0.0 | −0.45 | 3.14 | 3.02 | 0.066 | — |
| 0.1 | 1.38 | 3.73 | 3.15 | 0.604 | selects |
| 0.2 | 3.68 | 4.46 | 1.00 | 13.56 ✓ | selects |
| 0.4 | 11.67 | 8.15 | **−7.92** | **NA** | selects |
| 0.8 | 42.39 | 24.77 | **−23.30** | **NA** | selects |

`criterion = "wald"` selects the variable at every non-null effect size.

## Why it matters for 1.2.0

- Score is the **default** as of this release, and this is the first release where it is.
- A screen silently drops its best candidates. It is not fully silent since #117 — `$criteria$n_uncomputable_scores` counts them and the run warns — but the count does not say *which*, or that the reason is a large effect rather than a degenerate column.
- The production bagging screens read the reported counts as near zero, so it has not visibly bitten yet. That is not evidence it cannot: those runs used `%hazboot`-style pools where most candidates are weak.

## Directions, not decided

[...]

PR #135, opened 2026-08-19
fix: separate a candidate's own bad curvature from the model's

ehrlinger:
Closes the last actionable item of #130, and follows the source reading in [this comment](https://github.com/ehrlinger/temporal_hazard/issues/130#issuecomment-5342289997).

## The reference splits these; we did not

`src/vars/q1.c` in `ehrlinger/hazard`:

```c
diag = b1[n];
if (diag <= ZERO)  { *err = 2; return ZERO; }   /* flag 2 */
...
*qtol = ONE - DQDOT(wk, wk, n);
if (*qtol <= ZERO) { *err = 3; return ZERO; }   /* flag 3 */
```

Flag 2 is **the candidate's own observed information is not positive**. Flag 3 is **the candidate is unusable given what is already in the model**. R folded both into `information_indefinite`, whose user-facing text says the candidate is *"typically STRONG rather than degenerate"* and should be kept.

So a candidate whose own curvature is wrong was reported as one worth keeping — the same class of misdiagnosis as the sign-vs-magnitude defect in #132, pointing the other way.

## The branch is live, and censoring is what reaches it

I checked before writing the guard, because a guard for an unreachable state is dead code.

**Effect size does not reach it.** Swept the single candidate's true β on right-censored data:

| true β | 1 | 2 | 3 | 5 | 8 | 12 | 20 |
|---|---|---|---|---|---|---|---|
| `I_bb` | 41.9 | 101 | 120 | 133 | 131 | 129 | 127 |

Nor do the single-distribution families (weibull 138, loglogistic 49, lognormal 19, exponential 160).

**Censoring does.** A multiphase fit with 90% of its events interval-censored:

| interval share | 50% | 90% |
|---|---|---|
| `I_bb` at β=0.5 | **−0.0006** | **−0.003** |

Those rows contribute a difference of survival terms rather than a log density, so the candidate's own second derivative can go negative. That is the same corner of the package where the `Surv()` mistranslation, the silent Hessian and the missing score fallback all lived.

The fixture is therefore built from censoring, not effect size — and the test **asserts `I_bb <= 0` before asserting the reason**, so it cannot quietly stop exercising the branch and keep passing.

## Verification

Watched failing: `actual: "information_indefinite"` / `expected: "information_nonpositive"` against the parent commit.

`devtools::test()` **2095 pass / 0 fail**, `lintr::lint_package()` clean, `document()` a no-op.

`tol` drops its `abs()`, since `i_bb` is now known positive where it is formed.

## What is left on #130

Only item (3), the per-candidate Wald fallback — an enhancement, not a defect. Items (1) and (2) are closed: the reference uses observed information and R matches it, and the causes are now reported and split as the reference splits them.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,7 +37,7 @@ Suggests:
     roxygen2,
     rmarkdown,
     scales,
-    testthat (>= 3.0.0),
+    testthat (>= 3.2.0),
     withr
 LazyData: true
 Config/testthat/edition: 3
--- a/NEWS.md
+++ b/NEWS.md
@@ -102,7 +102,13 @@ selects.
 
   `$criteria$uncomputable_reasons` (and `$uncomputable_reasons` on a
   `mode = "select"` bootstrap) now counts the causes by name, `$all_scores`
-  carries a `reason` column per candidate, and both warnings name them. A run
+  carries a `reason` column per candidate, and both warnings name them. The
+  reference implementation separates these too, and the R side now matches its
+  split: a candidate whose *own* observed information is not positive is
+  reported apart from one that is unusable only given what is already in the
+  model (`information_nonpositive` against `collinear` and
+  `information_indefinite`). The first is reachable on a multiphase fit with a
+  large share of interval-censored rows. A run
   that *completed* while declining a candidate for this reason now warns too:
   it previously returned a selection -- sometimes an empty one -- in complete
   silence, which is the case where the omission is least visible. The
--- a/R/score-test.R
+++ b/R/score-test.R
@@ -643,8 +643,10 @@
   #                     far from zero -- a STRONG candidate -- and also when
   #                     `current` has not truly converged.
   #
-  # The magnitude test must come FIRST, and the sign test must be against
-  # -tol rather than 0. For an exactly collinear candidate the true v_beta is
+  # The magnitude test comes FIRST and the sign test is against -tol, not 0.
+  # The `-tol` is redundant given the early return above, and is written out
+  # anyway so that the rule holds on its own: reordering these two guards must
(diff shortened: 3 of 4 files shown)


ehrlinger:
All three review items addressed in `1f8a0d1`.

**1 & 2 — the `-tol` comment vs `v_beta < 0`.** Correct, and worth more than a comment edit. The two are equivalent only because `abs(v_beta) <= tol` returns first — the equivalence is a property of the guard *order*, and order is precisely what caused the defect this branch descends from ([#132](https://github.com/ehrlinger/temporal_hazard/pull/132)). Written out as `v_beta < -tol` so the rule stands on its own and a future reordering cannot quietly reintroduce it. No behaviour change: reaching that line already implies `abs(v_beta) > tol`.

**3 — the mock may not bind in the package namespace.** Checked rather than argued. Installed the package into a temp library and ran the file against the **installed** package, which is what `R CMD check` does:

```
R CMD INSTALL --library=/tmp/tlib .
Rscript -e '.libPaths("/tmp/tlib"); library(TemporalHazard); testthat::test_file(...)'
```

All 28 assertions report `success`, including the mocked branch at lines 177–178. The stub is effective under both `load_all()` and the installed package, so the concern does not reproduce. I made the change anyway — unqualified is the clearer way to say "the same namespace `local_mocked_bindings()` writes to", and it drops a `:::` from the suite.

**Also folded in:** `DESCRIPTION` declared `testthat (>= 3.0.0)` while `local_mocked_bindings()` needs **3.2.0** (introduced experimental in 3.1.7, announced in 3.2.0). `test-hessian-unavailable.R` has been calling it for some time, so the floor was already understated before this branch — this branch just adds the second caller. Raised to `>= 3.2.0`.

PR #137, opened 2026-08-19
fix: the multiphase derivatives disagreed with the likelihood on entry times

### LINKED ISSUE #136: hazard(): supplying time_lower/time_upper breaks a multiphase fit, even when they equal time
## Summary

Passing `time_lower` / `time_upper` to `hazard()` breaks a `dist = "multiphase"` fit **even when the bounds are identical to `time`**, which is a mathematical no-op. The optimizer then diverges to a flat plateau, returns a positive "objective" where a log-likelihood is expected, and reports `converged = TRUE` with a singular Hessian.

This is a silent-wrong-answer bug: nothing in the return value says the fit is meaningless except `rcond = 0`, and the accompanying warning blames the Hessian rather than the response.

## Reproducer

Public `avc` data shipped with the package; no covariates, no interval censoring, bounds deliberately set equal to `time` so they carry no information.

```r
library(TemporalHazard)
data("avc", package = "TemporalHazard")

ph  <- list(early    = hzr_phase("cdf", t_half = 0.19, nu = 1.44, m = 1, fixed = "m"),
            constant = hzr_phase("constant"))
th  <- c(log(0.35), log(0.19), 1.44, 1, log(0.03))
ctl <- list(n_starts = 1, maxit = 2000, conserve = TRUE)

a <- hazard(time = avc$int_dead, status = avc$dead,
            dist = "multiphase", phases = ph, theta = th, fit = TRUE, control = ctl)

b <- hazard(time = avc$int_dead, status = avc$dead,
            time_lower = avc$int_dead, time_upper = avc$int_dead,   # <- no-op bounds
            dist = "multiphase", phases = ph, theta = th, fit = TRUE, control = ctl)

a$fit$objective   # -210.5006   rcond 0.0518
b$fit$objective   #  49325.04   rcond 0        converged TRUE
```

## Expected

`b` should equal `a`. `time_lower` and `time_upper` both default to `time`, so supplying them explicitly at those values should change nothing.

## Actual

| call | objective | rcond | converged |
|---|---|---|---|
| no bounds | -210.5006 | 0.0518 | TRUE |
| bounds = `time` | 49325.04 | 0 | TRUE |

## How it shows up in real work

Found while reproducing a SAS `PROC HAZARD` interval-censored death model (SAS `time iu_dead; icensor ic_dead=il_dead;`, log-likelihood -239.194, `log10(condition) = 1.2258`, i.e. well conditioned). On that study, with genuine interval censoring:

- bounds supplied, `status` in `{0, 1, 2}` -> objective `50914.3488`, `rcond = 0`
- bounds supplied, `status` in `{0, 1}` -> objective `54450.0674`, `rcond = 0`
- **no bounds**, `status` in `{0, 1}` -> objective `-237.5761`, `rcond = 0.0673`, coefficients within ~0.2 of the SAS estimates

The objective was byte-identical (`50914.3488`) across genuinely different datasets - the full cohort, the cohort with the interval-censored rows dropped, and the cohort with them recoded - while the fitted coefficients ranged from -169 to +261 depending only on `control`. An objective that does not move when the data move, paired with coefficients that move freely, is the signature of a flat plateau: the response is not reaching the likelihood once bounds are present.

Starting the optimizer *at the known SAS optimum* did not help; it left immediately and landed on the same plateau.

## Suggested checks

1. Whether `time_lower` / `time_upper` are being bound into the multiphase objective at all, or are silently replacing the response.
2. A guard rejecting a "converged" multiphase fit whose objective is positive. The objective is a log-likelihood, so `+49325` is impossible on its face and would have caught this immediately.
3. `converged = TRUE` should not be returned alongside `rcond = 0`, or the two should be reconciled in `print`/`summary` output.

Related: `logLik()` has no method for class `hazard` (only `coef`, `predict`, `print`, `summary`, `vcov`), so the natural way to check a fitted log-likelihood is unavailable.

## Environment

- TemporalHazard 1.2.0
- R 4.6.1, aarch64-apple-darwin
- numDeriv 2016.8.1.1 installed, so the documented Suggests-related standard-error loss is not the cause

ehrlinger:
Closes #136 — though not in the way the report frames it. It found something worse than what it describes, and the thing it expected is not a defect.

## The real bug: the derivative was of a different function from the objective

For a `status` 0/1 row the log-likelihood subtracts `H(time_lower)` **unconditionally** whenever `time_lower` is supplied. The analytic gradient and Hessian defined the entry time with an extra `time_lower < time` filter, so a subject entering the risk set at its own event or censoring time was differentiated as though it had no entry time — while `has_start` still applied its weight. Three conditions for one concept, across two files.

Measured on `avc` at fixed θ, `max |analytic − numeric|`:

| `time_lower` | before | after |
|---|---|---|
| `NULL` | 3.5e−08 | 3.5e−08 |
| `= time` | **382** | 3.9e−09 |
| `= time − 0.5` (all rows) | 2.3e−09 | 2.3e−09 |
| **mixed** (some rows `== time`) | **126** | 3.0e−09 |

**The mixed row is the one that matters.** That is ordinary left-truncated data — some subjects enter at the moment they leave — not a pathological input, and it was silently wrong. Hessians likewise now agree to ~1e−6 across all five layouts.

## The reported expectation is wrong, and the docs are why

The report expects `b` (bounds `= time`) to equal `a` (no bounds). It should not, and it still does not after this fix. For `status` 0/1, `time_lower` is the counting-process **entry time**: setting it to `time` says every subject left the risk set at the instant it entered, so every such row drops out and the objective is unbounded above. The divergence is correct behaviour.

What produced the wrong expectation was the documentation:

> `time_lower` — Optional numeric lower bound vector for censoring intervals. Used when `status == 2` (interval-censored); **defaults to `time` if NULL.**

For `status` 0/1 the default entry time is **`0`**, not `time`. Read literally, the old wording says supplying `time_lower = time` changes nothing. Both roles are now documented, and `time_lower >= time` on a `status` 0/1 row warns with the count and the `NULL` default named:

```
310 of 310 row(s) have 'time_lower' >= 'time' with status 0 or 1. For those rows
'time_lower' is the counting-process entry time, so they enter the risk set at or
after they leave it and contribute nothing to the likelihood ... If you meant the
default -- entry at time 0 -- leave 'time_lower' as NULL; passing
'time_lower = time' is not the same thing.
```

Genuine left truncation stays silent — asserted, so the warning cannot become noise.

## On the issue's other suggestions

- *"Guard a converged fit whose objective is positive."* Not implemented. A log-likelihood of a continuous density can legitimately be positive, so a blanket sign test would false-positive. The invariant that actually catches this class is analytic-vs-`numDeriv`, which is now asserted.
- *`logLik()` has no `hazard` method.* True, and out of scope here — worth its own issue.

## Version

**1.2.1.** Nothing above 1.1.0 has shipped — no `v1.2.0` tag, no release — so this folds into the existing unreleased block rather than opening a NEWS section for a version nobody can install (the precedent from the 1.1.0 cycle). `DESCRIPTION`, `NEWS.md` and `cran-comments.md` headings all moved together.

## Verification

New `tests/testthat/test-time-lower-entry.R` asserts the analytic-vs-`numDeriv` invariant for gradient **and** Hessian across five entry-time layouts, rather than the one the old filter happened to admit. Watched failing: **7 of 19** against the parent commit.

`devtools::test()` **2107 pass / 0 fail**, `lintr::lint_package()` clean, `document()` regenerated `man/hazard.Rd`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: TemporalHazard
 Title: Temporal Parametric Hazard Modeling
-Version: 1.2.0
+Version: 1.2.1
 Authors@R: person("John", "Ehrlinger", email = "john.ehrlinger@gmail.com", role = c("aut", "cre", "cph"))
 URL: https://ehrlinger.github.io/temporal_hazard/, https://github.com/ehrlinger/temporal_hazard
 BugReports: https://github.com/ehrlinger/temporal_hazard/issues
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# TemporalHazard 1.2.0
+# TemporalHazard 1.2.1
 
 ## Breaking changes
 
@@ -86,6 +86,33 @@ selects.
   log-likelihood is not among them; that still comes from the `.lst`.
 
 ## Bug fixes
+
+* **The multiphase gradient and Hessian disagreed with the log-likelihood on
+  left-truncated data.** For a row with `status` in `{0, 1}` the log-likelihood
+  subtracts `H(time_lower)` unconditionally, but the analytic derivatives
+  defined the entry time with an extra `time_lower < time` filter. A subject
+  entering the risk set at its own event or censoring time was therefore
+  differentiated as though it had no entry time, while its weight was still
+  applied -- so the derivative was taken of a different function from the one
+  being evaluated, and the optimizer left any sensible region immediately.
+  Measured on `avc` at fixed parameters, the analytic gradient was out by 382
+  where every row entered at its exit time, and by 126 where only *some* did
+  -- which is ordinary left-truncated data, not a pathological input. Both
+  derivatives now define the entry time exactly as the likelihood does, and
+  new tests assert agreement with `numDeriv` across five entry-time layouts
+  rather than the one the old filter happened to admit.
+
+* **`hazard()` documented `time_lower` incorrectly, and now warns when it is
+  self-defeating.** The argument was described only as the lower bound of a
+  censoring interval, "defaulting to `time` if NULL". For `status` in
+  `{0, 1}` it is in fact the counting-process **entry time**, and leaving it
+  `NULL` means entry at `0`, *not* at `time`. Read literally, the old wording
+  said that passing `time_lower = time` changes nothing; it in fact states
+  that every subject left the risk set at the instant it entered, which
+  removes every such row from the likelihood and leaves the objective
+  unbounded above. The documentation now gives both roles, and supplying
+  `time_lower >= time` on a `status` 0 or 1 row warns, naming the count and
+  the `NULL` default. Reported as issue #136.
 
 * **`hzr_stepwise()` now says *why* a candidate could not be scored, and warns
   when the reason is that the candidate looks strong.** Under
--- a/R/hazard_api.R
+++ b/R/hazard_api.R
@@ -106,8 +106,18 @@ NULL
 #'
 #' @param time Numeric follow-up time vector.
 #' @param status Numeric or logical event indicator vector.
-#' @param time_lower Optional numeric lower bound vector for censoring intervals.
-#'   Used when `status == 2` (interval-censored); defaults to `time` if NULL.
+#' @param time_lower Optional numeric vector with two distinct roles, selected
+#'   by `status`. Supplying it explicitly is **not** a no-op.
+#'   * `status == 2` (interval-censored): the lower bound of the censoring
+#'     interval, defaulting to `time`.
+#'   * `status %in% c(0, 1)` (right-censored or event): the counting-process
+#'     **entry time**, so the row contributes `H(time) - H(time_lower)`.
+#'     Left `NULL`, the entry time is **`0`**, not `time`.
+#'
+#'   Passing `time_lower = time` therefore states that every subject entered
+#'   the risk set at the instant it left, which contributes nothing and
+#'   removes the row from the likelihood. That is a valid specification and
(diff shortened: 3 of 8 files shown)


ehrlinger:
Three of the four taken, one declined with evidence.

**Taken.**

- **`.entry_variants(k)` rebuilt every iteration** — correct, it ran ten times over a five-element list. Hoisted.
- **Hard-coded `310`** — correct, and the better kind of finding. The claim being asserted is that the warning counts *every* row, so it now derives from `length(k$time)`. A test that fails because a fixture grew is a test that gets edited rather than read.
- **"walked off a cliff"** — reworded, though not for the reason given. The problem is not that it is informal; this codebase's comments are deliberately direct. The problem is that it does not say *what happened*. Now: "the reported gradient pointed somewhere the objective did not, and the optimizer followed it out of the region where the fit means anything."

**Declined: `status %in% c(0, 1)` → `c(0L, 1L)`.**

The stated reason is consistency with the codebase, but the nearest neighbour in this same file is unsuffixed:

```r
# R/hazard_api.R:503
if (any(status %in% c(-1, 2))) {
```

The `0L`/`1L` idiom belongs to `R/likelihood-multiphase.R`, which operates on already-normalised status. `hazard_api.R` is the validation layer, reading what the user passed — where `status` may be double or integer, and `%in%` coerces either way. Matching the file beats matching a different file, so `c(0, 1)` stays.

Also merged `dev` in now that #135 has landed. `DESCRIPTION` carries both changes cleanly — `Version: 1.2.1` and `testthat (>= 3.2.0)`. Suite **2116 pass / 0 fail**, lintr clean.

Issue #134, opened 2026-08-19
stepwise accepts steps that lower the log-likelihood; score explodes on near-collinear candidates (mirror of #130)


ehrlinger:
Found in a production 2-phase forward selection (n = 3049, 251 candidates, `criterion = "score"`, `slentry = 0.10`, `max_steps = 10`) on TemporalHazard 1.2.0 at `dev@c105b9c6`.

This is the **accept-side mirror of #130**. That issue reports `v_beta` going negative so the score returns `NA` and a strong candidate is silently declined. Here `v_beta` collapses toward zero instead, `Q = u^2 / v` explodes, and a numerically degenerate candidate is accepted with `p = 0.000` while the refit that follows makes the model **worse**.

## Measured

| step | variable | phase | score `stat` (df 1) | logLik | delta |
|---|---|---|---|---|---|
| 6 | agee | late | 42.99 | -3498.241 | +22.086 |
| 7 | age2 | late | 612.20 | -3497.759 | +0.482 |
| 8 | ln_age | late | 1724.16 | -3503.771 | **-6.012** |
| 9 | age | late | **92210.90** | -3506.308 | **-2.537** |
| 10 | ln_ht | early | 1606.37 | -3554.932 | **-48.624** |

After step 6 the relationship inverts: **the larger the score statistic, the worse the outcome.** Step 6 scores 43 and gains 22 log-likelihood units. Step 9 scores 92,211 on 1 df, which is not a credible chi-square, and loses 2.5.

The final model has **19 coefficients and fits 57 log-likelihood units worse than the 16-coefficient model at step 7**. Those models are nested, so that is arithmetically impossible at the true maximum. It is proof of a failed refit, not a statistical result.

The context is four collinear forms of one concept accumulating in one phase (`agee`, `age2`, `ln_age`, `age`). As they do, the design goes near-singular, `.hzr_safe_solve()` emits `rcond = 0`, and the Schur complement stops being a usable variance.

## Two distinct defects

**A. The score statistic is unbounded when the candidate is near-collinear with the current model.**
Same root as #130: the Schur complement of the *observed* information is not a trustworthy variance. #130 catches the sign failure (`v_beta < 0` guarded to `NA`). There is no corresponding guard for `v_beta -> 0+`, which produces a finite but meaningless statistic that passes every downstream check. A `stat` of 92,211 on 1 df should not be reportable as evidence.

[...]

PR #140, opened 2026-08-19
fix: a forward step that lowers the log-likelihood is a failed refit

ehrlinger:
Defect **B** of #134. Defect A (the unbounded score statistic) is not in this PR — see the note at the end.

## The objective was written at every step and compared at none

A forward step enters a model that **contains** the one it started from, so at the optimum the log-likelihood cannot fall. `grep -n objective R/stepwise.R` returned three write sites and no read.

In the production screen that surfaced this, three of ten steps lowered the log-likelihood while the run reported `converged`, ten entries and `p = 0.000` throughout. The final 19-coefficient model fitted **57 units worse** than the nested 16-coefficient model three steps earlier — arithmetically impossible at a maximum. The only thing that noticed was a monotonicity check written by hand downstream, which every other caller would have had to reinvent.

## What changed

- `$steps` gains **`delta_logLik`**.
- A forward step that lowers the objective **warns**, naming the step, the drop, and *why* it is impossible — then is counted in `$criteria$n_nonmonotone_entries`.
- `hzr_bootstrap(scope = )` aggregates **`$n_nonmonotone_replicates`**, for the reason #117 and #132 both needed: a replicate whose path went backwards still contributes its selections to the pooled frequencies, indistinguishable from one that converged, and every replicate runs under `suppressWarnings()` so a step-level warning cannot surface in the mode that matters.
- The comparison carries a relative tolerance, so optimizer noise cannot fire it. The real cases are whole log-likelihood units, not `1e-10`.

## ⚠️ Two of the three tests first passed for the wrong reason

Worth stating, because it is the failure mode this package keeps finding. The mocked fixture subtracted a fixed amount from the **entered** objective. The genuine step gains ~14.6 units, so subtracting 5 left it **9.6 ahead of where it started** — the assertion held while the branch never fired, and the tolerance test never entered the tolerance band at all.

Caught by probing whether the mock fired, not by reading the green result. Both fixtures now set the objective relative to the model being entered *from*, which is the quantity the check actually compares.

## Verification

Watched failing: **6 of 8** against the parent commit. `devtools::test()` **2125 pass / 0 fail**, `lintr::lint_package()` clean, `document()` regenerated `man/hzr_bootstrap.Rd`.

## Not included

**Defect A** — the score statistic being unbounded as `v_beta -> 0+` (a `stat` of 92,211 on 1 df). #134 suggests expected rather than observed information, but the reference uses observed too, so that would break parity. The reference does have a guard we never implemented: `dqstat.c` bounds the *implied coefficient*, `if (fabs(*qbeta) > 50.0) { *qflag = 4; }`, calling it "the model is going to infinity". R computes `stat` and never derives `qbeta`. That is the parity-grounded fix for A, and it is a separate change.

B is deliberately first because it is independent of the criterion: it catches A whether or not A is resolved, and it catches any other failed refit too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -87,6 +87,25 @@ selects.
 
 ## Bug fixes
 
+* **`hzr_stepwise()` never checked that an accepted step improved the fit.**
+  A forward step enters a model that *contains* the one it started from, so at
+  the optimum the log-likelihood cannot fall. It was written into `$steps` at
+  every step and compared at none, so a step whose refit failed to converge
+  entered anyway and every later step was then scored against a model that was
+  not at its own optimum. In the production screen that surfaced this, three of
+  ten steps lowered the log-likelihood and the run still reported convergence,
+  ten entries and `p = 0.000` throughout; the final 19-coefficient model fitted
+  57 units worse than the nested 16-coefficient model from three steps earlier,
+  which cannot happen at a maximum.
+
+  `$steps` now carries `delta_logLik`, a forward step that lowers the objective
+  warns and is counted in `$criteria$n_nonmonotone_entries`, and
+  `hzr_bootstrap(scope = )` reports `$n_nonmonotone_replicates` — a replicate
+  whose path went backwards still contributes its selections to the pooled
+  frequencies, and each replicate runs under `suppressWarnings()` so the
+  step-level warning cannot reach the user. The comparison carries a small
+  tolerance so optimizer noise does not fire it. Reported as issue #134.
+
 * **The multiphase gradient and Hessian disagreed with the log-likelihood on
   left-truncated data.** For a row with `status` in `{0, 1}` the log-likelihood
   subtracts `H(time_lower)` unconditionally, but the analytic derivatives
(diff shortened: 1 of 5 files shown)


ehrlinger:
Good catch, and correct. Fixed in `9d3c1a2`.

The zero-row `$steps` fallback at `R/stepwise.R:459` is built by hand, separately from `record_step()` and `record_freeze()`, so the new column went into the row constructors and was missed there. `$steps` had **13 columns on a run that took no step and 14 on a run that took one**, and `r$steps$delta_logLik` was `NULL` rather than `numeric(0)` for the first. Added in the same position as in the row constructors, so column order matches as well.

**My own tests missed it** because all three took at least one step. The new test asserts the *whole schema* is identical between a run that stepped and one that did not:

```r
expect_identical(names(none$steps), names(taken$steps))
```

rather than asserting the one column exists. Presence-only would let the next column added to `$steps` repeat this exactly.

Watched failing: 3 of 14 with only the empty-frame line reverted. Suite 2130 pass / 0 fail, lintr clean.

PR #141, opened 2026-08-19
fix: decline a candidate whose implied coefficient runs to infinity

### LINKED ISSUE #134: stepwise accepts steps that lower the log-likelihood; score explodes on near-collinear candidates (mirror of #130)

ehrlinger:
Defect **A** of #134, following #140 (defect B). Independent of it — either can merge first.

## The cause is not what the issue assumed, and measuring that changed the fix

#134 proposes expected rather than observed information. But the reference uses **observed** information too ([established on #130](https://github.com/ehrlinger/temporal_hazard/issues/130#issuecomment-5342289997)), so switching would break parity.

More importantly, **near-collinearity alone does not explode `Q`**. Measured on `avc`, sweeping a candidate toward collinearity: `u_beta` shrinks along with `v_beta`, `Q` stays around 1–2, and the existing collinear guard then takes over. It never reaches the thousands.

`Q` explodes when **the model being scored against is not at its optimum** — the reduced-model score is then no longer zero, so the numerator is inflated while the denominator stays small. That is precisely the state a failed refit leaves behind, which is why #134 saw it at step 9, *after* steps 8 and 9 had lowered the log-likelihood.

Near-collinear candidate on `avc`, model displaced from its optimum:

| θ shift | u_β | v_β | Q | implied β | reason **before** |
|---|---|---|---|---|---|
| 0 | 0.102 | 0.0059 | 1.77 | 17.3 | — |
| 0.05 | −126.6 | 0.0062 | **2.6e6** | −20,326 | **—** |
| 0.25 | −704.7 | 0.0076 | **6.5e7** | −92,381 | **—** |

The 0.25 row brackets the reported 92,211. **Nothing was declined** — `v_beta` is positive and far above the collinearity floor in every row.

## The reference has guarded this all along

We had implemented three of `dqstat.c`'s four checks. The fourth forms the **implied coefficient** — one Newton step from β=0, `u_beta / v_beta` — and rejects it:

```c
*qbeta = (*qz)*(*qse);                 /* = u_beta / v_beta */
if (fabs(*qbeta) > 50.0e0) { *qflag = 4; return; }
```

> "the model is going to infinity ... legitimate for a variable that is either positive or negative with respect to all the remaining events in its phase. However the model cannot manage this."

A legitimate candidate reaches `|qbeta|` of about **14** in the same setting, so 50 has real headroom rather than clipping ordinary work. The new reason `coefficient_diverging` says both halves: the fit is running to infinity, *and* to check whether an earlier step's refit failed.

## Verification

Tested from the **real fixture**, not a stub, with the at-optimum case as the control — without it the test could not distinguish the guard from the fixture. A third test asserts the failure is specifically in the implied coefficient, not in `Q` being large: `v_beta > 0` and above the floor, while `|u/v| > 50`.

Watched failing: **4 of 13**. `devtools::test()` **2129 pass / 0 fail**, `lintr::lint_package()` clean, `document()` a no-op.

## Relationship to #140

Complementary rather than overlapping. #140 detects the failed refit *after* the step and warns; this declines the bad candidate *before* it is accepted. Since A's trigger is the state B detects, running both means a screen that hits this reports the cause twice — once as a diverging candidate, once as a step that lowered the log-likelihood.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -87,6 +87,25 @@ selects.
 
 ## Bug fixes
 
+* **A score statistic could be finite, enormous and meaningless.** A
+  production screen accepted a candidate with `stat` = 92,211 on 1 df and
+  `p = 0.000`, after which the refit made the model worse. Neither existing
+  guard reached it: the adjusted variance stayed positive and well above the
+  collinearity floor, so the statistic was reported as evidence.
+
+  Near-collinearity alone does not do this — as a candidate approaches
+  collinearity its score shrinks along with its variance and `Q` stays small.
+  `Q` explodes when the model being scored against is *not at its optimum*,
+  because the reduced-model score is then no longer zero: the numerator is
+  inflated while the denominator stays small. That is the state a failed refit
+  leaves behind. `hzr_stepwise()` now declines a candidate whose implied
+  coefficient exceeds ±50, reporting `coefficient_diverging`, which is what
+  the SAS/C reference has always done (`dqstat.c` rejects `|QBETA| > 50` as
+  "the model is going to infinity"). Measured on the bundled `avc` data, a
+  model displaced 0.25 from its optimum produced `Q` = 6.5e7 with no reason
+  reported at all; a legitimate candidate reaches an implied coefficient of
+  about 14, so the threshold has real headroom. Reported as issue #134.
+
 * **The multiphase gradient and Hessian disagreed with the log-likelihood on
   left-truncated data.** For a row with `status` in `{0, 1}` the log-likelihood
   subtracts `H(time_lower)` unconditionally, but the analytic derivatives
--- a/R/score-test.R
+++ b/R/score-test.R
@@ -665,6 +665,32 @@
     return(na_result("information_indefinite"))
   }
   stat <- (u_beta^2) / v_beta
+
+  # The coefficient this Q implies: one Newton step from beta = 0, with
+  # standard error 1 / sqrt(v_beta). SAS forms the same quantity and rejects
+  # the candidate when it is absurd -- `dqstat.c`:
+  #
+  #     *qz    = sqrt(q);
+  #     *qse   = -(*qz)/d1llad;
+  #     *qbeta = (*qz)*(*qse);          /* = u_beta / v_beta */
+  #     if (fabs(*qbeta) > 50.0e0) { *qflag = 4; return; }
+  #
+  # calling it "the model is going to infinity ... legitimate for a variable
+  # that is either positive or negative with respect to all the remaining
+  # events in its phase. However the model cannot manage this."
+  #
+  # This is the guard against a Q that is finite, enormous and meaningless.
+  # The other two guards do not reach it: v_beta stays positive and well
+  # above the collinearity floor. Measured on `avc` with a near-collinear
+  # candidate, scored against a model moved 0.25 off its optimum -- which is
+  # what a failed refit leaves behind -- Q was 6.5e7 with v_beta = 0.0076 and
+  # no reason reported at all. A legitimate candidate at the optimum reached
+  # |qbeta| of about 14 in the same setting, so 50 leaves real headroom.
+  q_beta <- u_beta / v_beta
+  if (!is.finite(q_beta) || abs(q_beta) > 50) {
+    return(na_result("coefficient_diverging"))
+  }
+
   list(stat = stat, df = 1L,
        p_value = stats::pchisq(stat, df = 1L, lower.tail = FALSE),
        reason = NA_character_)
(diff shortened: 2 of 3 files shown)


ehrlinger:
Merged `dev` in now that #140 has landed. **`NEWS.md` conflicted** and was resolved by hand: both PRs inserted a bullet immediately after `## Bug fixes`, so git could not choose. Both are additive and both are kept, #140's first to match the merge order. No other file conflicted.

Re-ran the gate on the merged tree: **2143 pass / 0 fail**, `lintr::lint_package()` clean, `document()` a no-op.

Worth noting the two guards now coexist as intended rather than overlapping. #140 detects a failed refit *after* a step lowers the log-likelihood; this PR declines the diverging candidate *before* it is accepted. A screen hitting the production pathology in #134 now reports the cause from both ends.

PR #171, opened 2026-08-23
docs(cran-comments): describe the 1.2.2 submission, measured not carried forward

ehrlinger:
`cran-comments.md` was still stamped **1.2.1** and described that release, while `DESCRIPTION` and the `NEWS.md` top heading are at 1.2.2. Nothing in this repo catches that — there is no version-grep test here.

## What was already right

The file's *scope*. Its Summary says "an update to TemporalHazard 1.1.0 (accepted 2026-06-12)" and its section heading is "Changes since 1.1.0" — and that is correct: `CRAN-SUBMISSION` records `Version: 1.1.0 / 2026-06-12`, and tags stop at `v1.1.0`. **1.2.0 and 1.2.1 were never submitted.** So this is not a 1.2.1 → 1.2.2 delta; it is 1.1.0 → 1.2.2, and the existing content stays.

## What this adds

**Summary** — one paragraph saying plainly that the work accumulated under two development version numbers that never went out, so `NEWS.md`'s 1.2.0 and 1.2.1 sections are all new to CRAN.

**New features** — `hazard()`'s vector interface evaluating in `data`'s scope (with the shadowing warning and the `data` type check), and `hzr_translate_sas()`, flagged experimental here because its own documentation flags it.

**Bug fixes** — a new **Multiphase fitting** group: `hzr_decompos()` at large `|m|`, RNG reproducibility and `control$start_seed`, the infeasible-shape start, `fit$fit$starts`, and `$se` conformability. The zero-step `hzr_stepwise()` screen joins the existing selection group.

## Measured, not carried forward

Every number in **Test environments** was re-measured on this tree. These are claims made to CRAN, so none of the previous values were assumed still true:

| Claim | Was | Measured |
|---|---|---|
| `R CMD check --as-cran` (with PDF manual) | `Status: OK` | **`Status: OK`** — 0/0/0 |
| Overall check time | "about three minutes" | **200s** (tests 79s, vignettes 41s) |
| Tarball size | 2.7 MB | **2.8 MB** (2,952,066 bytes) |
| `urlchecker::url_check()` | 19 URLs correct | **19 URLs, 0 flagged** |
| Reverse dependencies | 0 | **0** |

Built from a clean `git archive` export per `AGENTS.md`; `build/vignette.rds` is present and no `CLAUDE`/`AGENTS`/`cran-comments`/`RELEASE` files reach the tarball.

The aspell **NOTE disposition** section is unchanged and still accurate — `DESCRIPTION`'s `Description` text does not differ between `main` and `dev`.

## Gate

`cran-comments.md` is `.Rbuildignore`d (line 22), so this change cannot affect the tarball. The `R CMD check --as-cran` above ran on package content identical to this branch's. `document()`/`lintr` are not applicable to a build-ignored Markdown file and were not run — stating that rather than implying a gate that did not happen.

### DIFF
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -1,9 +1,13 @@
-# CRAN submission comments -- TemporalHazard 1.2.1
+# CRAN submission comments -- TemporalHazard 1.2.2
 
 ## Summary
 
 This is an update to TemporalHazard 1.1.0 (accepted 2026-06-12). There is no
 reviewer feedback to address.
+
+The work accumulated under two development version numbers, 1.2.0 and 1.2.1,
+that were never submitted. `NEWS.md` carries a section for each; everything in
+them is new to CRAN, and the changes below cover all of it.
 
 This is a minor-version release that nonetheless contains one breaking change:
 `hzr_stepwise()` now defaults to a different selection criterion, so re-running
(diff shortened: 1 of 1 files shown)


ehrlinger:
> `NEWS.md` has no 1.2.0 heading [...] As written, this tells CRAN to look for a section that does not exist.

Correct, and fixed in a7ebd7f. Verified against the file rather than taken on report:

```
$ grep -n '^# TemporalHazard' NEWS.md | head -3
1:# TemporalHazard 1.2.2
361:# TemporalHazard 1.2.1
739:# TemporalHazard 1.1.0
```

There is no 1.2.0 heading. The summary now says the work is recorded under the 1.2.2 and 1.2.1 headings, with 1.2.0's folded into the 1.2.1 section rather than kept separate, which is what actually happened.

I did not add the missing section, the other option offered. 1.2.0 was never submitted to CRAN, so folding it into 1.2.1 is a reasonable record of a release that did not happen, and splitting it back out now would mean reconstructing a boundary that no tag or `CRAN-SUBMISSION` entry marks. Worth a maintainer decision rather than a silent fix: anyone who installed from GitHub `main` sees `Version: 1.2.0` in `DESCRIPTION` against a NEWS heading that no longer exists.

PR #172, opened 2026-08-23
release: TemporalHazard 1.2.2

ehrlinger:
Release PR for **TemporalHazard 1.2.2** — 259 commits, ~113 files since `main`.

## Read this first: 1.1.0 is what is on CRAN

`CRAN-SUBMISSION` records `Version: 1.1.0 / 2026-06-12`, and the tag list stops at `v1.1.0`. Under the tag-after-acceptance policy in `RELEASE.md`, that means **neither 1.2.0 nor 1.2.1 was ever submitted**. Two consequences:

1. This release ships the accumulated work of three version numbers. `NEWS.md` records it under its 1.2.2 and 1.2.1 headings (1.2.0's work was folded into the 1.2.1 section rather than kept separate), and `cran-comments.md` is scoped "Changes since 1.1.0" accordingly.
2. `main` currently carries `Version: 1.2.0` without a release behind it — it got there via `68e4645 chore(release): target 1.2.0 for the next CRAN drop`, which is the "do not pre-stamp a future release number" anti-pattern `RELEASE.md` documents. Merging this restores the invariant that `main` names a real release.

**This release includes one breaking change**, inherited from the unreleased 1.2.1: `hzr_stepwise()` now defaults to `criterion = "score"`, so re-running an existing stepwise analysis can select a different variable set. `criterion = "wald"` restores the old behaviour exactly. It is flagged under a "Breaking changes" heading in `NEWS.md` and in `?hzr_stepwise`.

## Release gate

Measured on `dev` at **`8eebb31`** — the tree this PR merges — from a tarball built out of a clean `git archive` export per `AGENTS.md`:

| Step | Result |
|---|---|
| `R CMD check --as-cran`, **with the PDF manual** | **`Status: OK`** — 0 errors, 0 warnings, 0 notes (exit 0) |
| Overall check time | **212s** (3m 32s) — inside CRAN's ~10 min ceiling |
| Tests (under check) | 81s / 86s · FAIL 0, SKIP 146, PASS 2342 |
| Vignette rebuild | 42s / 46s |
| Tarball | **2.8 MB** (2,956,816 bytes), `build/vignette.rds` present |
| Tarball hygiene | no `CLAUDE`/`AGENTS`/`cran-comments`/`RELEASE` files |
| Local suite (`NOT_CRAN=true`) | FAIL 0 / SKIP 6 / **PASS 3196** |
| `lintr::lint_package()` | **0** |
| `urlchecker::url_check()` | 19 URLs, **0 flagged** |
| Reverse dependencies | **0** |

Note the gap between the two test lines, which is the point `AGENTS.md` makes about not reading a green check as full coverage: **146 skips / 2342 passes** under check against **6 skips / 3196 passes** locally. The `skip_on_cran()` tests are not evidenced by CI.

The check budget in `AGENTS.md` was re-measured and re-stamped this cycle (#176) after it had drifted from 2m 52s to ~3m 30s unnoticed. This run at 3m 32s agrees with that entry within noise, so the recorded baseline is current.

## Dependency changes

New `Suggests`: **`haven`**, **`withr`**. `testthat` raised to **`>= 3.2.0`**.

## Review

The `RELEASE.md` step 5 adversarial review of the accumulated `main...dev` diff was run and its findings are [posted as a comment](https://github.com/ehrlinger/temporal_hazard/pull/172#issuecomment-5387672649), split into what was confirmed by execution, what was relayed unverified, and what the review did not cover.

Four findings from it were fixed and merged into `dev` before this gate: #173 (`$se` discarding computable standard errors), #174 (the vignette showing an emitted document the translator no longer produces), #175 (the INHAZ remedy not being valid R), #176 (the stale check budget). Two are **open and deliberately not fixed** — the `scope = NULL` logical-column abort, and four relayed findings listed in that comment.

## Not done here

Per `RELEASE.md`, the tag and GitHub Release come **after CRAN acceptance**, not with this merge — if CRAN asks for changes you need another commit on `main`, and a pre-acceptance tag has to be moved. Remaining sequence after merge: `devtools::check_win_devel()` (the source of truth for the aspell NOTE, which a local `--as-cran` under-reports), `devtools::submit_cran()`, then on acceptance `git tag -a v1.2.2` and the GitHub Release.


ehrlinger:
## `RELEASE.md` step 5 — adversarial review of the accumulated `main...dev` diff

Run over the full 1.2.2 release diff (244 commits, ~113 files, ~22.9k insertions). Step 5 is **advisory**: findings are triage input, not a pass/fail, and each was verified against the code rather than taken on report. This comment marks explicitly which findings were confirmed **by execution**, which are relayed unverified, and what the review did not cover — a silent cap reads as full coverage.

### Confirmed by execution

**1. `$se` is one `NA` per parameter, not one standard error.** `R/hazard_api.R:652` returned `rep(NA_real_, ncol(vcov_mat))` whenever `anyNA(vcov_mat)`. A multiphase fit with a fixed shape — the normal case the fix targeted — measured on `avc`:

```
$se        : NA  NA  NA  NA  NA
diag(vcov) : 0.02238  0.4367  0.1914  NA  0.2985
summary SE : 0.1496  0.6609  0.4375  NA  0.5464
```

Four computable standard errors discarded because one entry is `NA`, and `$se` contradicting `summary()` on the same object. This is the hollow-object shape from `AGENTS.md`: right length, empty inside, defeating both `is.null()` and a length check. #170 made `$se` conformable but not correct, and both `NEWS.md` and `cran-comments.md` asserted "one standard error per parameter, whatever the variance matrix looks like" — false as written.

Being fixed in a follow-up PR: elementwise computation, the `NEWS.md` claim corrected, and the existing regression test strengthened. That test asserted only `expect_length` and `expect_named`, both of which a vector of `NA` satisfies — it passed throughout the bug. It now asserts the fixture reaches both branches, that computable SEs survive, and that `$se` equals `summary()`'s `std_error`. Reverting the fix fails 2 assertions, so it can fail.

**2. `scope = NULL` offers logical columns, then aborts the screen on them.** `.hzr_modellable_vars()` (`R/stepwise-formula.R:316`) deliberately keeps logicals — its roxygen says "Logical columns count: they are ordinary 0/1 predictors" — and both criteria then refuse them before step 1, violating the filter's own stated contract that under `scope = NULL` dropping beats erroring. The two paths fail for *different* reasons:

```
score -> candidate 'mal_lgl' is not numeric (logical)...
wald  -> Unknown coefficient name(s): 'constant.mal_lgl'.
         Available: 'constant.log_mu', 'constant.mal_lglTRUE'
```

The `wald` model fits the logical fine as `mal_lglTRUE`; that half is `.hzr_candidate_coef_name()` not accounting for the factor-style `TRUE` suffix, which is narrower and more fixable than "logicals unsupported". Not fixed here.

**3. The recorded check budget is stale.** `AGENTS.md` records overall 2m52s with tests at 53s/58s. Two independent runs on this tree: 3m14s / tests 76–79s, and 200s / tests 79s. Tests are up roughly 45%. Still well inside CRAN's ~10 min ceiling, but `AGENTS.md`'s figures are the early-warning signal a future agent will trust, and they no longer describe this package.

### Relayed, not independently verified

- `hzr_bootstrap()` on a vector-interface fit with no `data =` fails naming the wrong cause (`R/diagnostics.R:1447`, `n_obs <- nrow(orig_data)` is `NULL`).
- `vignettes/sas-to-r-migration.qmd:571` shows an emitted document with `status = DEAD` and no `weights` — which this release's own NEWS says changed. Prose contradicting code is the failure mode `AGENTS.md` names.
- `$call_env` (`R/hazard_api.R:728`) doubles serialized fit size when `data` is bound in a non-global frame, and breaks `identical()` on two fits from the same call. Additive, but an observable change to the returned shape that NEWS does not mention.
- `nu = 0` with `-h < m < 0` still probes `m + h > 0` and errors out of `hzr_decompos()` (`R/hessian-multiphase.R:56`). Degrades loudly to numDeriv, so a lost analytic Hessian rather than a wrong number.
- `NEWS.md:345` says four start statuses; the code assigns five, and the roxygen correctly lists five.

### Not covered

`R/sas-parse-job.R` and the rest of the SAS lexer/parser beyond roxygen and one end-to-end run; `R/sysdata.rda` contents; ~7,600 lines of new tests; `R/candidate-score.R` and the stepwise drop path; G3 late-phase second derivatives. The NEWS corpus claims ("57 of 110 files translate into 22 documents", "grid resolution 19 of 55") are **unverified** — they need the private `hazard` corpus, which is not in this checkout.

### Verified sound

The `hzr_decompos()` log-scale rewrite was re-derived by hand and reconfirmed against the pre-rewrite formulas over a 250-point grid, with `G(t_half) = 0.5` to 15 digits at large `|m|` and every numeric claim in NEWS reproduced. The `Surv()` → package status translation is correct for all four forms. The score-test assembly is consistent, with Q = 11.37 / Wald = 10.44 / LRT = 9.89 for `mal` on `avc` — the expected ordering.

ehrlinger:
> The unresolved-INHAZ message embeds an example `librefs = c(<lib> = "<path>")` but `<lib>` is inserted unquoted.

Valid, and fixed in #174's sibling PR rather than here, since this PR is the `dev` → `main` release merge and takes no direct commits.

One narrowing after reading the code. SAS librefs are constrained to letter-or-underscore-initial alphanumerics of at most 8 characters, so they are almost always syntactic R names and the emitted `c(EX = "<path>")` parses fine. The failure needs a libref colliding with an R **reserved word** — `TRUE`, `if` — which is legal in SAS and a syntax error in R. So this is narrower than "if the libref is not a syntactic R name", but it is real, and since the line is advertised as copy-pasteable, quoting is unconditionally correct.

Fixed by quoting the libref, with a test that **runs** the advertised remedy rather than matching its text: it renders a job with `INHAZ=TRUE.HZD`, extracts the `c(...)` fragment from the emitted message, `parse()`s and evaluates it, and asserts the result is named `TRUE`. Reverting to the unquoted form fails that test.

Note the existing `test-sas-render-qmd.R:43` covers a related-but-different case, an apostrophe in the libref breaking the *chunk*. This one is about the remedy *inside* the message being runnable.

PR #173, opened 2026-08-23
fix(hazard): $se keeps the standard errors it can compute

ehrlinger:
Follow-up to #170, found by the `RELEASE.md` step 5 review of the accumulated 1.2.2 diff (posted on #172).

## What #170 missed

#170 made `$se` conformable with `$par` and stopped there — any `NA` anywhere in the variance matrix returned one `NA` per parameter:

```r
if (anyNA(vcov_mat)) {
  return(rep(NA_real_, ncol(vcov_mat)))
}
```

A multiphase fit holding a shape fixed carries `NA` variance rows **by design** — that is the case #170 was written for — so in practice this returned a vector of the right length and empty of everything it should carry. Measured on an `avc` two-phase fit with `m` fixed:

```
$se        : NA  NA  NA  NA  NA
diag(vcov) : 0.02238  0.4367  0.1914  NA  0.2985
summary SE : 0.1496  0.6609  0.4375  NA  0.5464
```

Four computable standard errors discarded because one entry was `NA`, and `$se` contradicting `summary()` on the same object. That is the hollow-object shape in `AGENTS.md`'s table: right length, nothing inside, defeats both `is.null()` and a length check.

## The fix

`.hzr_safe_se_from_vcov()` computes element by element. A parameter whose variance is missing or invalid gets `NA`; every other parameter keeps its own standard error.

This also removes the `any(d < 0)` guard, which had the identical all-or-nothing flaw — one small negative variance from a numerical Hessian discarded every good one beside it. A scalar `NA` still means there is no variance matrix at all, the contract `vcov.hazard()` keeps.

Nothing internal consumes `$se` (`wald.R` and `predict-cl.R` go to `vcov` directly), so the blast radius is the fit object's own field.

## The test was the real problem

The #170 regression test asserted only `expect_length()` and `expect_named()` — and `rep(NA_real_, 5)` satisfies both. It passed for the entire life of the bug. Strengthened in place rather than duplicated:

- assert the fixture reaches **both** branches before comparing anything (`any(computable)` and `any(!computable)`) — coverage first, per `AGENTS.md`;
- assert the computable standard errors survive and equal `sqrt` of their variances;
- assert `$se` equals `summary()`'s `std_error`, which is the invariant that actually broke.

**Mutation-tested:** reinstating the `anyNA` early return fails 2 assertions.

## NEWS

The 1.2.2 entry claimed "one standard error per parameter, whatever the variance matrix looks like", which was true only in the sizing sense. Corrected to describe the elementwise behaviour, and to say why sizing alone is not enough. `cran-comments.md` carried the same overstatement and is corrected in #171.

## Gate

`document()` clean (no `man/` churn) · `lintr::lint_package()` **0** · `devtools::test()` **0 FAIL / 6 SKIP / 3192 PASS** (`NOT_CRAN=true`)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -203,10 +203,14 @@
   the vector [1]".
 
   A scalar `NA` now carries the meaning it already has for `vcov()`, that there
-  is no variance matrix at all. Where a matrix is present, `$se` is sized to it.
-  `summary()` was never affected, since it builds its coefficient table from
-  the variance matrix directly, so this was a quiet inconsistency on the fit
-  object rather than a visible break.
+  is no variance matrix at all. Where a matrix is present, `$se` is computed
+  element by element: a parameter held fixed carries an NA variance row and
+  earns an `NA` standard error, while every parameter whose variance *was*
+  computed keeps its own. Sizing the vector to the matrix is not enough on its
+  own -- filling it with `NA` throughout would leave `$se` conformable and
+  empty, and contradicting `summary()`, which reads the same matrix and reports
+  those standard errors. `summary()` was never affected either way, so this was
+  a quiet inconsistency on the fit object rather than a visible break.
 
 * `hzr_decompos()` no longer returns a wrong value for large `|m|`. The three
   branches with a nonzero `m` all formed `2^m` and the terms built from it
--- a/R/hazard_api.R
+++ b/R/hazard_api.R
@@ -657,6 +657,9 @@ hazard <- function(formula = NULL,
     # discarded standard errors that summary() reports from the same matrix,
     # leaving `$se` the right length and empty of everything it should carry.
     out <- rep(NA_real_, length(d))
+    # rep() starts unnamed; carry over whatever diag() gave us so a named vcov
+    # still yields a named `$se`, as the old sqrt(d) path did.
+    names(out) <- names(d)
     ok <- is.finite(d) & d >= 0
     out[ok] <- sqrt(d[ok])
     out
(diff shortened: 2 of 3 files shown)


ehrlinger:
> `.hzr_safe_se_from_vcov()` now builds `out` with `rep()`, which drops the diagonal names from `d <- diag(vcov_mat)`. [...] Preserve `names(d)` on the output vector.

Right in principle, and taken. `names(out) <- names(d)` before the assignment, so a named variance matrix still yields a named `$se` exactly as the old `sqrt(d)` path did.

Worth recording that I could not reproduce the regression, though, because it changes how the new assertion should be read. No fitted path I tried produces a named `fit$fit$vcov`:

```
weibull      fit$fit$vcov dimnames: FALSE   names(se): NULL
exponential  fit$fit$vcov dimnames: FALSE   names(se): NULL
loglogistic  fit$fit$vcov dimnames: FALSE   names(se): NULL
multiphase   fit$fit$vcov dimnames: FALSE   names(se): NULL
```

So `sqrt(d)` was not naming anything either, and no caller can currently be relying on `names(fit$fit$se)`. (`vcov.hazard()` does name its result, at `R/hazard_api.R:1609`, but that is a labelled copy built by the S3 method, not the `$vcov` field this helper reads.) The path that makes a named diagonal reachable is `R/hessian-invert.R:64`, `dimnames(vcov) <- dimnames(H)`, which fires only when the Hessian itself carries dimnames.

The test therefore asserts `names(fit$fit$se)` equals `names(diag(fit$fit$vcov))`, which holds trivially today with both `NULL`. That is a guard against the two diverging later, not a reproduction of a fault, and the comment above it says so rather than letting a passing assertion imply more coverage than it has.

Gate after the change: `lintr` 0, `devtools::test()` 0 FAIL / 6 SKIP / **3193 PASS**.

PR #174, opened 2026-08-23
docs(vignette): the emitted-document example matches what the translator emits

ehrlinger:
Found by the `RELEASE.md` step 5 review of the 1.2.2 diff (see #172). Verified by running the translator on `hz-example.sas`, the file the vignette itself cites.

## The drift

| `sas-to-r-migration.qmd` said | `hzr_translate_sas()` actually emits |
|---|---|
| "carries a `data` chunk and a `fit` chunk" | `setup`, `data`, **`status`**, `fit` |
| `status = DEAD` | `status = .hzr_status` |
| *(no `weights`)* | `weights = ifelse(DEAD > 0, DEAD, 1)` |

That contradicts this release's own changelog. `NEWS.md` states that "an `EVENT` count carries into `weights` and `status` derives from `EVENT > 0`", and `hz-example.sas` has `EVENT DEAD;`, so it takes exactly that path.

**No gate could have caught this.** Both blocks are `#| eval: false`, so the vignette rebuild renders them as literal text — `R CMD check` verifies the vignette *builds*, never that its displayed output matches what the code produces.

## Why it matters here

This vignette is written for the bilingual SAS-to-R migrant, the reader most likely to lift that emitted `hazard()` call as a template. Carrying `status = DEAD` into a real fit is not a cosmetic error: before 1.2.2, `DEAD = 2` mapped onto `status = 2` and was fitted as **interval-censored** — a different likelihood branch, not an under-count.

## What changed

- The narrative names all three chunks and points at `status` as the one to read closely.
- The shown output is what the translator emits, including the `status` chunk and `weights`.
- Two new paragraphs: why one SAS `EVENT` statement becomes a `status`/`weights` pair, and why the guard refuses rows SAS would have deleted (`readc1.c` sets `mdel` on a missing count and `del` on a negative one; `readobs.c` skips the row and subtracts it from `Nobs`, so it contributes nothing to the likelihood). Translating such a row as censored would move the fit without saying so.

The guard's full `stop()` text is trimmed in the displayed chunk, marked as trimmed, and explained in the prose beneath it.

## Gate

`lintr::lint_package()` **0** · `devtools::spell_check()` clean · `R CMD build` renders the vignette. No R code changed.

### DIFF
--- a/vignettes/sas-to-r-migration.qmd
+++ b/vignettes/sas-to-r-migration.qmd
@@ -568,12 +568,15 @@ job <- hzr_translate_sas(
 )
 ```
 
-The emitted `.qmd` carries a `data` chunk and a `fit` chunk (reformatted
-here for width; the emitted document wraps at 60 columns like any
-`deparse()`d call). `DATA` step translation is out of scope, so the
-`data` chunk is a guard, not a re-implementation -- it checks that
-`AVCS` is already in scope and stops with a named error if not, rather
-than silently proceeding without data:
+The emitted `.qmd` carries three chunks past its `setup`: a `data`
+chunk, a `status` chunk, and a `fit` chunk (reformatted here for width;
+the emitted document wraps at 60 columns like any `deparse()`d call).
+`DATA` step translation is out of scope, so the `data` chunk is a
+guard, not a re-implementation. It checks that `AVCS` is already in
+scope and stops with a named error if not, rather than silently
+proceeding without data. The `status` chunk is where `EVENT DEAD`
+becomes something `hazard()` can take, and it is worth reading closely
+if you are checking a translation against the job it came from:
 
 ```{r}
 #| eval: false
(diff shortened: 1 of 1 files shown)


ehrlinger:
> The prose says the emitted `.qmd` has separate `data`, `status`, and `fit` chunks, but the example shown below is a single Quarto chunk [...] This reads like a contradiction.

Agreed, and fixed in a055ea7. I took the first of the two options offered, in substance: rather than only stating the concatenation in prose, the block now marks each boundary where it falls.

```r
#| label: translate-output
# chunk: data
if (!exists("AVCS")) {
...
# chunk: status
AVCS <- transform(AVCS, .hzr_status = {
...
# chunk: fit
fit <- hazard(
```

The sentence above it now says the three are run together in one block, each marked by a comment, rather than shown as the three separate chunks they are. Showing them as three literal Quarto chunks would mean nesting fences inside a displayed chunk, which reads worse than the markers do.

Worth recording that the ambiguity predates this branch: the block already showed two emitted chunks as one while the prose called them two. Adding the third made it conspicuous rather than introducing it.

This matters more than a formatting nit for this vignette specifically. Its reader is the SAS migrant checking a translation against the job it came from, so the boundaries are exactly what they need to line up against their own emitted file.

Re-gated: `lintr` 0, `devtools::spell_check()` clean. The change is comments inside an `eval: false` block, so it cannot affect rendering; CI's `R CMD check` rebuilds the vignettes across the matrix and will confirm.

PR #175, opened 2026-08-23
fix(sas): the unresolved-INHAZ remedy is R you can paste

ehrlinger:
Addresses the Copilot review item on #172. That PR is the `dev` → `main` release merge and takes no direct commits, so the fix lands here.

## The item, narrowed

`R/sas-render-qmd.R:38` built the remedy with the libref inserted bare:

```r
"unresolved INHAZ=", job$inhaz, " -- pass librefs = c(", lib,
' = "<path>") to hzr_translate_sas()'
```

Copilot framed this as "if the libref is not a syntactic R name". After reading the code that is wider than the real exposure: SAS librefs are letter-or-underscore initial, at most 8 alphanumerics, and `lib` has already had everything past the first `.` stripped — so a libref is almost always a syntactic R name and `c(EX = "<path>")` parses fine.

The case that does break is a libref colliding with an R **reserved word**. SAS accepts one named `TRUE`; `c(TRUE = "<path>")` is a syntax error. Narrow, but the line exists to be pasted, and quoting is unconditionally correct.

## The test runs the remedy

Rather than asserting the message's shape, the test executes what it advertises: render a job with `INHAZ=TRUE.HZD`, pull the `c(...)` fragment out of the emitted message, `parse()` and evaluate it, assert the result is named `TRUE`.

**Mutation-tested:** restoring the unquoted form fails it (1 FAIL).

That choice is deliberate. A string-matching assertion here would pass on `c(TRUE = "<path>")` — the exact broken output — because it *looks* right. The bug is only visible if you run it.

`test-sas-render-qmd.R:43` already covers a neighbouring case, an apostrophe in the libref breaking the *chunk*. This is the separate question of whether the remedy *inside* the message is runnable.

## Gate

`document()` clean · `lintr::lint_package()` **0** · `devtools::test()` **0 FAIL / 6 SKIP / 3190 PASS** (`NOT_CRAN=true`)

### DIFF
--- a/R/sas-render-qmd.R
+++ b/R/sas-render-qmd.R
@@ -32,9 +32,12 @@
 
   if (!is.null(job$inhaz) && !isTRUE(job$inhaz_resolved)) {
     lib <- sub("[.].*$", "", job$inhaz)
+    # Quote the libref. A SAS libref is nearly always a syntactic R name, but
+    # one that collides with a reserved word (TRUE, if) would make the remedy
+    # invalid R, and the whole point of the line is that it can be pasted.
     msg <- paste0(
-      "unresolved INHAZ=", job$inhaz, " -- pass librefs = c(", lib,
-      ' = "<path>") to hzr_translate_sas()'
+      "unresolved INHAZ=", job$inhaz, ' -- pass librefs = c("', lib,
+      '" = "<path>") to hzr_translate_sas()'
     )
     stop_call <- bquote(stop(.(msg)))
     add("```{r}",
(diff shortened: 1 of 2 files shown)


ehrlinger:
> The regex uses a non-greedy quantifier `.*?` but `sub()` is called without `perl = TRUE`, which can be invalid or behave inconsistently depending on the regex engine.

Change taken in 34288a4, on narrower grounds than stated.

R's default regex engine is TRE, which supports non-greedy quantifiers as a documented extension, so `sub()` without `perl = TRUE` handles `.*?` here. Both forms return byte-identical results:

```
nongreedy   : c("TRUE" = "<path>")
first-paren : c("TRUE" = "<path>")
identical   : TRUE
```

So the test was not invalid and was not behaving inconsistently. What is true is that it leaned on an engine extension for no benefit: the fragment has no nested parentheses, so `[^)]*` says the same thing and needs nothing beyond POSIX ERE. That is worth having on its own, so the change is in.

Re-gated after the edit, including the mutation check, since a regex change is exactly the kind of edit that can quietly stop a test from failing when it should: `lintr` 0, `sas-render-qmd` 15 PASS, and restoring the unquoted libref still fails it (1 FAIL). The kill is unchanged.

PR #177, opened 2026-08-24
docs: drop the dev branch, work one line on main

ehrlinger:
Drops the `main` + `dev` split for a single line on `main`.

## Why

A second long-lived branch earns its keep when a stable line must keep shipping patches while a diverged line runs in parallel. TemporalHazard has no such line, so the split bought nothing and cost a back-merge after every `main`-side fix.

ggRandomForests **does** have that constraint — greenfield RHF work alongside a maintained CRAN release — and keeps its feature branch. The two repos differ here on purpose, and `RELEASE.md` now says so.

The premise the split rested on had also stopped holding. `RELEASE.md` claimed `main` "always == the latest version on CRAN". In fact 1.2.0 and 1.2.1 were both numbered and never submitted, and `v1.2.2` is tagged as a GitHub release with no CRAN submission behind it.

## Two things this fixes outright

- **`Closes #123` now works.** GitHub auto-closes an issue only on merge into the *default* branch. Under the split, routine work merged to `dev`, so the keyword silently did nothing and every issue had to be closed by hand — `AGENTS.md` carried a ⚠️ warning about it. That warning is now inverted rather than deleted, so the history stays legible.
- **The release-PR trap is gone.** With "Automatically delete head branches" on, merging a `dev -> main` release PR deletes `dev`, because `dev` is that PR's head branch. That happened on #172 this morning. Nothing was lost — `main` already contained every commit — but `dev` had to be pushed back from `main`. With one line there is no release PR and no head branch to delete.

## Changes

| File | What |
|---|---|
| `RELEASE.md` | Branch model rewritten to one line. Release flow, versioning cycle, step 5 review diff, step 9 submit checkout, and the acceptance steps re-anchored on `main`. Dated entry added to the historical note. |
| `AGENTS.md` | "Never push to `main` or `dev`" → "Never push to `main`". `Closes #123` warning inverted. Two measurement stamps now cite `c9f6c28` rather than "`dev` at `c9f6c28`". |
| `.github/workflows/` | `dev` removed from push/pull_request triggers in `R-CMD-check`, `lint`, `pkgdown`, `test-coverage`. |

## One design decision worth flagging

`RELEASE.md` step 5 ran the adversarial release review over `git diff main...dev`. With `dev` gone it needs a new anchor, and it now uses **`git diff v<previous>...main`**.

That is arguably better than what it replaced: a tag is an immovable record of exactly what shipped, whereas `dev`'s position depended on when you looked at it.

## Verification

`lintr::lint_package()` **0**. All seven workflow files still parse, with triggers confirmed:

```
R-CMD-check   -> push [main, master]  pull_request [main, master]
lint          -> push [main, master]  pull_request [main, master]
pkgdown       -> push [main, master]  pull_request [main, master]
test-coverage -> push [main, master]  pull_request [main, master]
check-manual  -> push [main]  release  workflow_dispatch
check-release -> release  workflow_dispatch
spelling      -> push [main]  pull_request
```

`c9f6c28` was verified to be an ancestor of `main`, so the re-worded measurement stamps still resolve after `dev` goes away.

No R code changed.

## Not done here

The remote `dev` branch still exists. Deleting it is a separate, destructive step — I have left it for you to confirm.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -9,11 +9,11 @@ on:
   push:
     paths-ignore:
       - '.claude/**'
-    branches: [main, master, dev]
+    branches: [main, master]
   pull_request:
     paths-ignore:
       - '.claude/**'
-    branches: [main, master, dev]
+    branches: [main, master]
 
 concurrency:
   # One in-flight run per workflow per ref. Without this every push starts a
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -2,9 +2,9 @@ name: lint
 
 on:
   push:
-    branches: [main, master, dev]
+    branches: [main, master]
   pull_request:
-    branches: [main, master, dev]
+    branches: [main, master]
 
 concurrency:
   # One in-flight run per workflow per ref. Without this every push starts a
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -9,11 +9,11 @@ on:
   push:
     paths-ignore:
       - '.claude/**'
-    branches: [main, master, dev]
+    branches: [main, master]
   pull_request:
     paths-ignore:
       - '.claude/**'
-    branches: [main, master, dev]
+    branches: [main, master]
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -9,11 +9,11 @@ on:
   push:
     paths-ignore:
       - '.claude/**'
-    branches: [main, master, dev]
+    branches: [main, master]
   pull_request:
     paths-ignore:
       - '.claude/**'
-    branches: [main, master, dev]
+    branches: [main, master]
 
 concurrency:
   # One in-flight run per workflow per ref. Without this every push starts a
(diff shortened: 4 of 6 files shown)


ehrlinger:
> `v<previous>` uses shell redirection syntax (`<` / `>`), so copying this command verbatim can fail or write to an unintended file.

Confirmed and fixed in the latest commit — with one correction to the second half.

Verified by running it in an empty directory:

```
$ git diff v<previous>...main
(eval):3: no such file or directory: previous
$ ls -a
.  ..
```

It fails exactly as described: the shell reads `<previous` as input redirection and errors before `git` is invoked. But **nothing is written**. The shell aborts on the missing input file before it ever reaches `>...main`, so no stray file appears. "Can fail" is right; "write to an unintended file" does not occur here.

Fixed in both files by dropping the placeholder for prose plus a concrete, runnable example:

```
git diff v1.2.2...main
```

which is the actual command to run today, and is copy-pasteable as written.

### Two pre-existing instances, deliberately not touched

The same pattern predates this PR, in `RELEASE.md` release-ops commands:

- `:183` — `git tag -a vX.Y.Z -m "... <date>" <main-HEAD>`
- `:195` — `gh release create ... --notes-file <news-section.md>`

Both would break the same way if pasted (`<main-HEAD>` in particular parses as a redirect from `main-HEAD` followed by a dangling `>`). They are outside this PR's scope, which is the branch model, so I have left them rather than widening the diff. Happy to fold them in here or send a separate PR — say which you prefer.

`lintr::lint_package()` remains 0.

PR #179, opened 2026-08-24
fix(stepwise): a logical candidate screens as the 0/1 predictor it is

ehrlinger:
The one genuine open defect from the `RELEASE.md` step 5 review of the 1.2.2 diff (see #172). Reproduced by execution before and after.

## The contradiction

`.hzr_modellable_vars()` offers logical columns as candidates under `scope = NULL`, and its roxygen says why:

> Logical columns count: they are ordinary 0/1 predictors, and whether a 0/1 field arrives logical or numeric depends on the reader that produced the frame rather than on anything about the variable.

Both criteria then refused what the package had just offered itself:

```
score -> candidate 'mal' in phase 'constant' is not numeric (logical)
wald  -> Unknown coefficient name(s): 'constant.mal'.
         Available: 'constant.log_mu', 'constant.malTRUE'
```

The screen stopped **before step 1**, on a column no caller had named. That also violates the filter's own stated contract, that under `scope = NULL` dropping beats erroring.

## Two independent defects

**1. The score path refused `is.logical` outright**, at three sites (`.hzr_score_check_numeric` plus two guards in `score-test.R`). A new `.hzr_candidate_numeric()` sits beside `.hzr_modellable_vars()` — the function that declares logicals modellable — and coerces once, so the refit sees the same column the screen scored.

**2. `.hzr_candidate_coef_name()` discarded the answer it had found.** When the guessed name is absent from `coef()`, it greps for the expansion and errors if there is more than one — then `return(target)` regardless, throwing away a single match:

```r
      if (length(expanded) > 1L) {
        stop(...)
      }
    }
    return(target)          # <- the one match found above is dropped
```

**This is not logical-specific.** `model.matrix()` renames any single-column expansion — a logical becomes `varTRUE`, a two-level factor `varb` — so a two-level factor was broken the same way and is fixed with it. A candidate expanding to more than one column is still refused, unchanged.

## The test asserts the claim, not the absence of an error

That a screen "runs" is the weak version. What is actually being claimed is that a logical column *is* its numeric twin, so the test compares two runs on identical data differing only in that column's type:

- same variables entered, in the same order (`steps$variable`, `steps$action`)
- on the same evidence (`steps$p_value`, tolerance 1e-6)
- and the same fit underneath (`coef()`, tolerance 1e-6) — names differ by design, so values are compared

It also asserts the logical column **actually enters** (`expect_true("mal" %in% steps$variable)`). Without that, two screens that both ignored it would agree, and the comparison would pass while proving nothing.

**Mutation-tested per half**, since one fix could mask the other:

| Mutant | Result |
|---|---|
| restore the numeric-only score guard | FAIL 1 — the `score` test only |
| discard the single expansion again | FAIL 1 — the `wald` test only |

Each kills its own criterion's test and not the other's.

## Version

Patch bump to **1.2.3**, `v1.2.2` being tagged; `NEWS.md` heading kept in sync by hand, since nothing in this repo greps for it. Minor/major untouched.

## Gate

`document()` clean (no `man/` churn — `@noRd` internals) · `lintr::lint_package()` **0** · `devtools::spell_check()` clean (`modelled` added to `inst/WORDLIST`, matching the existing `behaviour`/`recognise`) · `devtools::test()` **0 FAIL / 6 SKIP / 3214 PASS**

The suite was re-run after `document()` was fixed — an earlier `document()` failure meant the first run executed against a tree whose `man/`/`NAMESPACE` had not been regenerated.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: TemporalHazard
 Title: Temporal Parametric Hazard Modeling
-Version: 1.2.2
+Version: 1.2.3
 Authors@R: person("John", "Ehrlinger", email = "john.ehrlinger@gmail.com", role = c("aut", "cre", "cph"))
 URL: https://ehrlinger.github.io/temporal_hazard/, https://github.com/ehrlinger/temporal_hazard
 BugReports: https://github.com/ehrlinger/temporal_hazard/issues
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,11 @@
   expands the same way (`varb`). A candidate that expands to more than one
   column is still refused, unchanged.
 
+  That makes `criterion = "wald"` a real answer for a two-level factor or
+  character column named in an explicit `scope`, which the score criterion
+  still cannot expand. Its refusal used to say switching criterion would not
+  help, and now points at it instead.
+
 # TemporalHazard 1.2.2
 
 ## New features
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,25 @@
+# TemporalHazard 1.2.3
+
+## Bug fixes
+
+* A logical column no longer stops a stepwise screen dead. `hzr_stepwise(scope
+  = NULL)` enumerates its own candidates and counts logical columns among them,
+  on the grounds that a 0/1 field arriving logical rather than numeric is a
+  property of the reader that produced the frame, not of the variable. Both
+  criteria then refused what the package had offered: the score criterion
+  errored with "is not numeric (logical)", and the Wald criterion failed
+  looking up `phase.var` when `model.matrix()` had named the column
+  `phase.varTRUE`. Either way the screen stopped before its first step, on a
+  column nobody had chosen by hand.
+
+  A logical candidate is now modelled as the 0/1 predictor it is, and gives
+  the same screen as the identical numeric column: same variables entered, in
+  the same order, at the same p-values, with the same coefficients.
+
+  The coefficient-name half of this also fixes a two-level factor, which
+  expands the same way (`varb`). A candidate that expands to more than one
+  column is still refused, unchanged.
+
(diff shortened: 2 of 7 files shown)


ehrlinger:
> The score-path non-numeric error message still claims switching to `criterion = "wald"` won't help, which is now misleading for single-column expansions in the multiphase path.

Correct, and the sharpest catch of this review round — **this branch is what falsified it.** Fixed in 22898dc.

The claim was true when written. Once `.hzr_candidate_coef_name()` returns the expansion it finds, the Wald path handles a term expanding to a single design-matrix column, so the score path's refusal started steering callers away from the one option that works. Measured, on an explicit scope naming a two-level factor:

```
score -> ERROR: candidate 'mal' in phase 'constant' is not numeric (factor) ...
wald  -> OK - entered: mal,age
```

### Boundary checked, not assumed

The new wording had to avoid overclaiming in the other direction, so I measured where Wald actually stops:

| Input, explicit scope | Wald |
|---|---|
| two-level factor | **works** |
| two-level character | **works** |
| three-level factor | errors — "expands to multiple coefficients" |

So the message now says a two-level factor or character column works under Wald, and that anything wider needs expanding into numeric main effects.

### Why no existing test caught it

The message is only reachable through an **explicit** `scope`. Under `scope = NULL`, `.hzr_modellable_vars()` filters factors out before they reach it — which is exactly why the reproduction for the logical fix never touched this path.

### The test pins both halves

Asserting the message changed is not enough: the remedy it names has to work, or naming it is worse than naming none.

- the refusal no longer contains "switching criterion will not help"
- it does name `criterion = "wald"`
- and Wald actually enters `mal` on that same input

**Mutation-tested:** restoring the old wording fails it (1 FAIL).

`document()` clean · `lintr` 0 · `spell_check()` clean · `devtools::test()` **0 FAIL / 6 SKIP / 3218 PASS**

PR #194, opened 2026-08-31
feat(diagnostics): name the parameters when a fit sits on a likelihood ridge

ehrlinger:
Rescues the one unique commit from `claude/competent-pare-7518ac`, a branch cut against the
deleted `origin/dev` and stranded since. This is the findings half of the 1.2.2 cycle; the
release shipped without it.

## Why only one commit

That branch carried two. The first, `31f2821` ("stop losing an optimization start to an
infeasible shape"), is **already on `main`** as `10822c9`, rebased in via #166. Rebasing the
branch would have replayed it against itself. Only `1863a43` is unique, so only `1863a43` is
cherry-picked here.

The branch's *finding* — that the multiphase gradient fixture has no interior optimum in `m`
— also already landed, as prose, in `7a1a904`. What was missing is the detector.

## What this adds

An ill-conditioned Hessian already warned that standard errors were unreliable. That
understates a ridge: where the likelihood is near-flat along a parameter combination, the
individual point estimates along it are not determined by the data either — only the
combination is. The fit still reported `converged` with an ordinary-looking coefficient
table, so nothing on the object signalled it.

`.hzr_weak_direction()` (new, in `R/hessian-invert.R`) reads the flat direction and:

- `hazard()` warns once per fit, naming the spanning parameters with their correlation and `rcond`;
- `summary()` prints the same note;
- the finding is recorded as `fit$weak` (`NULL` when well identified) so it can be checked
  rather than scraped from a warning text.

Two design points worth review:

**The direction comes from the correlation of the estimates, not the raw covariance.**
Parameters sit on very different scales — `m ~ 27` against `nu ~ 0.027` — and a raw-covariance
eigenvector loads 1.000 on `m` alone, which would report this two-parameter ridge as one
unidentified parameter. The correlation form splits it 0.487/0.481.

**A merely imprecise parameter is deliberately not reported.** It must clear
`.hzr_ridge_cor_tol` (0.99), and the whole check is gated on the *existing* `.hzr_rcond_tol`,
so it never fires where the `rcond` warning is already silent.

Generic across distributions — nothing in it knows about phase shapes.

## Conflicts resolved

- `R/hazard_api.R` `@return`: `main` grew a richer multi-start `status` vocabulary
  (`"nonconverged"`, `"infeasible"`, the `maxit` note) after this branch was cut. Kept `main`'s
  text verbatim and appended the `weak` component. The incoming version had also placed `weak`
  inside the *multiphase-`starts`* parenthetical, which is wrong — `weak` is a component of
  `fit` and applies to every distribution — so it is described there instead.
- `man/hazard.Rd`: generated, regenerated by `document()`.
- `NEWS.md` merged without a conflict but placed the entry in the **1.2.2** section. Moved to
  a new 1.2.7 section; verified the move is purely additive (zero removed lines against
  `origin/main`).

## Gate

Run in full, in a clean worktree off `origin/main`:

- `devtools::document()` — clean
- `lintr::lint_package()` — **0 lints**
- `devtools::test()` with `NOT_CRAN=true` — **FAIL 0 | SKIP 6 | PASS 3341**
- `R CMD check --as-cran` with the manual, from a `git archive` export of the committed tree —
  **Status: OK**, 3m30s overall (tests 86s/91s, vignettes 47s/50s), tarball 2.9 MB. Confirmed
  no `CLAUDE`/`AGENTS` files in the tarball.

Version bumped 1.2.6 → **1.2.7** (patch; `DESCRIPTION` and the `NEWS.md` heading kept in sync
by hand, since nothing here greps for it).

> [!NOTE]
> The companion PR re-landing #130 also opens a 1.2.7 `NEWS.md` section. Whichever merges
> second will take a trivial `NEWS.md` conflict under the shared heading.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: TemporalHazard
 Title: Temporal Parametric Hazard Modeling
-Version: 1.2.6
+Version: 1.2.7
 Authors@R: person("John", "Ehrlinger", email = "john.ehrlinger@gmail.com", role = c("aut", "cre", "cph"))
 URL: https://ehrlinger.github.io/TemporalHazard/, https://github.com/ehrlinger/TemporalHazard
 BugReports: https://github.com/ehrlinger/TemporalHazard/issues
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,37 @@
+# TemporalHazard 1.2.7
+
+## New features
+
+* **A fit sitting on a likelihood ridge now says so, and names the
+  parameters.** An ill-conditioned Hessian already warned that standard
+  errors were unreliable. That understates the problem when the
+  ill-conditioning is a ridge: the likelihood is near-flat along some
+  combination of parameters, and there the individual *point estimates* are
+  not determined by the data either -- only the combination is. The fit still
+  reports `converged`, and the coefficient table still prints a number for
+  every parameter, so nothing on the object signalled it.
+
+  `hazard()` now warns, once per fit, naming the parameters that span the flat
+  direction along with their correlation and the Hessian's `rcond`, and
+  `summary()` prints the same note. The finding is recorded on the object as
+  `fit$weak` (`NULL` when the fit is well identified), so it can be checked
+  programmatically rather than scraped from a warning.
+
+  The direction is read off the *correlation* of the estimates rather than
+  their raw covariance. Parameters here sit on very different scales -- an `m`
+  of 27 against a `nu` of 0.027 -- and in raw units a direction that moves
+  both equally in statistical terms loads almost entirely on the larger one,
+  which would report a two-parameter ridge as a single unidentified parameter.
+  A parameter that is merely imprecise, without trading off against another,
+  is deliberately not reported: that is ordinary low precision, and the
+  existing `rcond` warning and the parameter's own standard error already
+  cover it.
+
+  The check is generic -- it runs for every distribution and knows nothing
+  about phase shapes -- and is gated on the `rcond` threshold the package
+  already uses, so it never fires where the ill-conditioning warning stays
+  silent.
+
 # TemporalHazard 1.2.6
 
 ## Bug fixes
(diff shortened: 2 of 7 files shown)


ehrlinger:
Review pass (r-reviewer) found three defects, each reproduced before being fixed; pushed as `396fcb9`. Details in the commit message. Remainder filed as #197.

**Silent false negative (high).** `.hzr_weak_direction()` took the leading eigenvector and *then* applied the correlation gate, so a real ridge was missed whenever a larger block of moderately correlated parameters carried more variance — an equicorrelated block of k parameters has eigenvalue `1 + (k-1)r`, passing a perfect two-parameter ridge's ceiling of 2 as soon as `r > 1/(k-1)`. Measured on a 5-parameter fixture with a ridge at r = 0.99999: correct at rho = 0.45, **NULL at 0.51, 0.60 and 0.75** — and `NULL` is the documented signal for *well identified*. This branch's own multiphase fixture sits 0.5% from tripping it (ridge eigenvalue 2.0225 against a 1.6587 runner-up from the `log_mu` pair). Now scans directions flattest-first and returns the first genuine trade-off.

**Unnamed parameters (high).** `optim()` returns an unnamed `par` for the single-distribution fits, so the warning read `par1`/`par2` while `summary()` printed the real names against the same numbers — four of five distributions, and `NEWS.md` claimed otherwise. Names now come from `.hzr_parameter_names()`, the same source `summary()` uses.

**A test that could not fail (high).** The only end-to-end negative test called `hazard(fit = TRUE)` with no `theta`; the fit branch is `else if (fit && !is.null(theta))`, so nothing was optimised and the detector returned at its first guard. Both assertions passed with the whole function deleted. It now passes `theta` and asserts the fit reached the detector first.

An eigenvalue floor was tried and **removed**: 20,000 random correlation matrices produced no case where it changed the answer, so it was a guard no test could distinguish — the test went with it rather than shipping an assertion that cannot fail.

Three mutations checked and killed. Gate re-run in full: 0 lints, **FAIL 0 | SKIP 6 | PASS 3356**, `R CMD check --as-cran` **Status: OK**. `signalled` added to `inst/WORDLIST` — the spelling job caught it, and a local `--as-cran` does not run aspell unless installed.

PR #207, opened 2026-09-01
chore: recompose the house style artifact

ehrlinger:
`.claude/house-style.md` had gone stale against its vault sources, so the
`house-style` CI job fails. **It fails on `main` too** — verified by running
the composer's `--check` against a clean `main` — so this was blocking every
open PR rather than being caused by any of them.

The vault source `r-package-structure.md` gained a section, *"The bump is not
part of the pull request"*, which is why the hash moved:

```
-    r-package-structure.md         sha256:c43a528bedf9
+    r-package-structure.md         sha256:cc35f4c2e152
```

Regenerated with `compose-house-style.R --repo TemporalHazard`, not hand-edited:
the file carries a `DO NOT EDIT` banner, and per `AGENTS.md` the next recompose
would revert any manual change.

Nothing else is touched. `+70 / −3`, all inside the generated artifact.

> **Follow-up, not done here:** the new section says a pull request lands
> without touching `Version:`, and that `NEWS.md` carries a standing level-one
> `# TemporalHazard (unreleased)` heading that every PR writes under. `NEWS.md`
> currently opens with `# TemporalHazard 1.2.8`, which is unreleased and
> untagged. Renaming that to `(unreleased)` is the maintainer's call about
> whether 1.2.8 is a named set, so I have left it. My remaining open PRs
> already do not bump `Version:`.

### DIFF
--- a/.claude/house-style.md
+++ b/.claude/house-style.md
@@ -13,7 +13,7 @@
     writing-voice.md               sha256:15e75ca9cb97
     writing-reader-profile.md      sha256:5131ade189c9
     writing-context.md             sha256:51f197dc0c97
-    r-package-structure.md         sha256:c43a528bedf9
+    r-package-structure.md         sha256:cc35f4c2e152
 -->
 
 # House Style — TemporalHazard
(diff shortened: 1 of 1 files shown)


ehrlinger:
Superseded. The recomposed `.claude/house-style.md` landed in #205 instead, because a `.claude/`-only pull request could not be merged at all: `R-CMD-check` and `spelling` carry `paths-ignore: '.claude/**'` while main's ruleset requires those contexts, so five required checks would never have run here.

#209 fixes that deadlock. This branch is now behind main and its diff would revert #205's changes, so it is closed rather than merged.

Issue #160, opened 2026-08-22
Translate SELECTION into a working hzr_stepwise() call (needs a formula-interface base fit)


ehrlinger:
Split out of #152. The translator now **refuses** `SELECTION` rather than emitting a call that cannot screen; this issue tracks making it work.

## Why it was refused

A translated `SELECTION` job needs `hzr_stepwise()`, and `hzr_stepwise()`'s refit path requires a **formula-interface** base fit (`R/stepwise-refit.R:88-91`). The translator emits the **vector** interface, because SAS's censoring statements map onto this package's `-1/0/1/2` status coding, which `survival::Surv()` does not share.

With a vector-interface base fit every candidate refit fails, and the forward step downgrades those failures to warnings and returns zero steps — a screen that renders clean and selects nothing. See the companion issue on that indistinguishability.

## What making it work requires

1. **A formula-interface base fit for `SELECTION` jobs**, i.e. emitting `Surv(...) ~ 1` instead of `time = , status = `. That means translating this package's status coding into `Surv()`'s different integers for the same meanings — the exact translation `AGENTS.md` records as a shipped wrong-answer bug (`Surv(type = "left")` read left-censored rows as *right*-censored). It needs verifying against the reference implementation, not just written.
2. **A phase-keyed `scope`.** `.hzr_stepwise_candidates()` requires a named list keyed by phase (`R/stepwise-formula.R:159`), which is also the correct meaning: SAS's `EARLY ... / CONSTANT ... / LATE ...` is a per-phase candidate pool, not a flat one. `tests/testthat/test-sas-parity.R:498` has the canonical form.
3. **The candidate pool must not be forced in.** Today the translator bakes `EARLY X1, X2, X3` into the phase formula as model terms, which inverts the statement's meaning. The fix is to withhold the covariates where they are gathered (`R/sas-parse-job.R:217-263`, before the `.hzr_parse_parms()` call) and use their names for `scope`.
4. **Its own chunk.** `fit <- hzr_stepwise(fit = fit, ...)` is self-referential; the stepwise call belongs in a chunk after the fit chunk, retargeted via `.hzr_retarget_fit()`.

## Scale

Measured across three production studies: **26** jobs use `SELECTION` with `PROC HAZARD`. Of those, **25 also use `ICENSOR`** and only **1** is plain `EVENT`/`TIME`. So a fix limited to the plain case would cover a single job — which is why the whole construct is refused for now rather than partially supported.

[...]

Issue #159, opened 2026-08-22
hzr_stepwise() cannot distinguish 'every refit failed' from 'nothing met slentry'


ehrlinger:
Found while investigating #152. This is a defect in `hzr_stepwise()` itself, not in the SAS translator.

## The defect

`.hzr_refit_with_scope()` (`R/stepwise-refit.R:88-91`) requires a formula-interface base fit:

```r
raw_formula <- current$call$formula
if (is.null(raw_formula)) {
  stop("`current` was not built via the formula interface; refit ",
       "requires a formula-based base fit.", call. = FALSE)
}
```

The forward step catches candidate-refit failures and downgrades them to warnings (`R/stepwise-step.R:210`, `:227`). So when **every** refit fails — for example because the base fit came from the vector interface — the run completes, emits warnings, and returns **zero steps**.

Zero steps is exactly what a screen returns when no candidate met `slentry`. The two outcomes are indistinguishable in the returned object.

## Why this matters

`AGENTS.md` already lists this failure shape in its table of shipped defects:

> A stepwise screen that stopped cleanly — it could not *score* any candidate, which is indistinguishable from finding none good enough.

`hzr_stepwise()` tracks that distinction for uncomputable **scores** (`information_indefinite`), but not for failed **refits**. The same reasoning that motivated the score-side flag applies here.

## Failure scenario

A caller passes a vector-interface base fit to `hzr_stepwise()`. Every candidate refit errors, each becoming a warning. The function returns a zero-row `$steps` and a model identical to the base fit. A reader sees a screen that ran and selected nothing, and concludes no covariate met the entry criterion — when in fact nothing was ever evaluated.

## Fix sketch

Record refit failures the way score failures are recorded, so a zero-step result can state *why* it is empty. Consider erroring outright when **no** candidate could be refit at all, since that is a broken call rather than a null result.

Relatedly, `hzr_stepwise()` could reject a vector-interface base fit up front with a clear message, rather than letting every candidate fail one at a time.

## Provenance

Surfaced during the 1.2.2 translator work: the planned `SELECTION` fix would have emitted a `hzr_stepwise()` call against a vector-interface base fit, producing exactly this silent no-op on 25 of 26 production `SELECTION` jobs. The translator now refuses `SELECTION` instead (#152), so this is not currently reachable through the translator — but it is reachable by any hand-written call.

ehrlinger:
[...]

PR #210, opened 2026-09-02
feat(stepwise): let a multiphase model refit from a vector-interface base fit

ehrlinger:
Requirement 1 of #160: `hzr_stepwise()` can now take a **multiphase** base fit
built through the vector interface (`time =` / `status =`), not only through
the formula interface.

## Why the restriction was incidental

`.hzr_refit_blocker()` (#159) refused every fit with no stored formula, for
every distribution. That is right for a single distribution: a scope change
there adds or drops a term in the **global** formula, so without one there is
nothing to mutate. It is wrong for multiphase: a multiphase scope change
rewrites the **phase** formula, and the global formula only ever carried the
response. Such a fit refits perfectly well from its stored response vectors.

The cost of the blanket refusal was every translated SAS `SELECTION` job. SAS's
censoring statements map onto this package's `-1/0/1/2` coding, which
`survival::Surv()` does not share, and 25 of the 26 such jobs also use
`ICENSOR`. Requiring a formula there would have forced exactly the status-code
round-trip `AGENTS.md` records as having shipped a wrong answer.

## What changed

- `.hzr_refit_blocker()` is distribution-aware **without a signature change**:
  it reads `fit$spec$dist` itself, so both callers still share one predicate
  and cannot drift, which was the stated reason for sharing it in #159.
- It also refuses a multiphase vector fit with no stored `time`/`status`,
  which would otherwise reach `hazard()` as a missing argument several frames
  down.
- The refit hands back `time_lower` / `time_upper` explicitly. The formula path
  gets those from `Surv()`; a vector refit that dropped them would silently
  refit a left-truncated cohort as at risk from time 0: a full result, no
  error, wrong answer. `expect_equal(out$data$time_lower, D$entry)` is the test
  that stands between the code and that failure mode.
- The single-distribution path keeps erroring, deliberately, and now asserts
  rather than letting a `NULL` formula travel into `.hzr_formula_update()`.
- `hzr_stepwise()`'s pre-flight remedy is distribution-specific: telling a
  multiphase caller to "rebuild with the formula interface" would be wrong
  twice over.

## Note on the re-land

This is **not** a cherry-pick of the earlier `f8e8a56`. `main` added
`.hzr_refit_blocker()` upstream of the lines that commit changed, so a
mechanical re-land would have merged cleanly and been **inert**. It is
rewritten against main's guard, and rebased onto `602e294`.

## Gate

Run at `015f6d3`, rebased onto `main` @ `602e294`:

- `devtools::document()`: clean, no `man/` / `NAMESPACE` drift
- `lintr::lint_package()`: **0 lints**
- `devtools::test()` with `NOT_CRAN=true`: **0 FAIL / 6 SKIP / 3551 PASS**
  (the 6 skips are the SAS fixture-availability baseline)
- `test-stepwise-vector-interface.R` alone: **0 FAIL / 0 SKIP / 23 PASS**,
  run separately to confirm its 8 `skip_on_cran()` tests genuinely execute
- `R CMD check --as-cran` with the manual, built from a clean `git archive`
  export of the committed tree, **Status: OK** (0 errors, 0 warnings,
  0 notes). Tarball 2.9 MB; tests 107s, vignette rebuild 52s.

## Not in this PR

No `NEWS.md` entry: #160 gets **one** bullet when requirements 2–5 land.
Requirements 2-5, namely phase-keyed scope, withholding the candidate pool from the
phase formulas, `hzr_stepwise()` in its own chunk via `.hzr_retarget_fit()`,
and dropping the translator refusal in `R/sas-parse-job.R`, are not started.

Part of #160.

### DIFF
--- a/R/stepwise-refit.R
+++ b/R/stepwise-refit.R
@@ -39,6 +39,18 @@
 #' (#159). Both callers read this one function so the two answers cannot
 #' drift apart.
 #'
+#' The answer differs by distribution, and deliberately so. A
+#' single-distribution scope change adds or drops a term in the **global**
+#' formula, so without one there is nothing to mutate. A multiphase scope
+#' change rewrites the **phase** formula instead, and the global formula only
+#' ever carried the response --- so a vector-interface multiphase fit refits
+#' perfectly well from its stored response vectors (#160). Blocking it shut
+#' out every translated SAS `SELECTION` job, since SAS's censoring statements
+#' map onto this package's `-1/0/1/2` coding, which `survival::Surv()` does
+#' not share; 25 of the 26 such jobs also use `ICENSOR`. Requiring a formula
+#' there would have forced exactly the status-code round-trip `AGENTS.md`
+#' records as having shipped a wrong-answer bug.
+#'
 #' @param fit A fitted `hazard` object.
 #' @return `NULL` when the fit can be refit, otherwise a character scalar
 #'   naming the obstruction, phrased to follow "... because".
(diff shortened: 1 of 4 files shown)


ehrlinger:
Addressed the review comment; thanks, the gap was real.

Taking the suggestion literally would not have closed it, though. The existing test's fixture is `ev = rbinom(n, 1, 0.75)` (status 0/1 only) with `D$entry <- 0`, and passes no `time_upper`, so `out$data$time_upper` is `NULL` whether the refit carries the field through or drops it. `expect_equal(out$data$time_upper, ...)` there is an assertion that cannot fail.

22fc3e7 adds a fixture where both bounds are load bearing instead: interval-censored rows (status `2`) with genuinely wide intervals, plus a non-zero entry time on the remaining rows. Guard-the-guard assertions run first so a fixture that degenerated to zero entry times or zero-width intervals fails loudly rather than making the bound assertions vacuous.

Mutation-tested: removing `"time_upper"` from the carried fields fails this test and nothing else in the file (FAIL 1 / PASS 26). File is now 0 FAIL / 0 SKIP / 30 PASS.

Issue #216, opened 2026-09-02
test-weights.R: weighted/duplicated parity is asserted over unidentified coefficients


ehrlinger:
Found by the `RELEASE.md` step-5 adversarial review of `git diff v1.2.2...main`, verified at `602e294`.

## What happens

`tests/testthat/test-weights.R:690-715` asserts weighted-vs-duplicated-row parity:

```r
expect_equal(coef(fit_w), coef(fit_dup), tolerance = 1e-3)
```

Both fits emit the phase-identifiability warning (observed in a full `NOT_CRAN=true` suite run, once per fit):

> Phase 'early' contributes at most 2.69e-09 of the cumulative hazard at any observed time. Such a phase has not started by the end of follow-up, so neither its 'mu' nor its shape is identified: the fit converges and those parameters drift freely.

The assertion covers the **whole** coefficient vector, `early.log_mu` and `early.x` included. Those are free-drifting in both fits. The comparison holds only because both take the identical deterministic path from the identical start (`n_starts = 1L`), not because the weighted and duplicated-row likelihoods agree on them.

## Why it matters

Part of what the parity claim rests on is agreement between undetermined quantities. Any change to the optimizer's path (a different start, a tolerance, a BLAS) could move both fits and break the test for reasons unrelated to weighting, or mask a genuine weighting defect in the identified parameters.

Note this is the **new identifiability guard doing its job**: it found a real problem on first contact with the existing suite. The test is `skip_on_cran()`, and CI sets `NOT_CRAN: true`, so these warnings now appear in the logs on all five platforms.

## Suggested fix

Either re-tune the fixture (`make_mp_cov(seed = 43)` / the `t_half = 0.3` start) so `early` is alive over the observed window, or restrict the comparison to the identified parameters and say in a comment why the others are excluded.

Issue #215, opened 2026-09-02
test-stepwise-integration.R: 'scope = NULL' test passes over a screen that scored nothing


ehrlinger:
Found by the `RELEASE.md` step-5 adversarial review of `git diff v1.2.2...main`. **Reproduced independently at `602e294`**, with a correction to the review's own account (see below).

## What happens

`tests/testthat/test-stepwise-integration.R:316-329`, *"scope = NULL works when the base formula was passed by variable"*, passes over a screen that could not score a single candidate.

Reproduced directly:

```
steps rows: 1
uncomputable reasons: nuisance_singular 6
warning: Stepwise selection stopped because the score statistic could not be
         computed for any remaining candidate (6 candidate scores were NA ...)
does expect_false(any(c("int_dead","dead") %in% sw$steps$variable)) pass?  TRUE
```

The test's only assertions are `expect_s3_class(sw, "hazard")` and `expect_false(any(c("int_dead", "dead") %in% sw$steps$variable))`. Since **nothing entered**, no variable name can appear, so the second assertion passes without discriminating anything.

> **Correction to the review:** it reported `$steps` as having zero rows and the reason as `information_indefinite`. Verified: `$steps` has **1** row and the reason code is **`nuisance_singular`** (×6). The conclusion is unchanged (the screen scored nothing), but the mechanism is a different one.

## Why it matters

This is row five of the `AGENTS.md` table verbatim: *"A stepwise screen that stopped cleanly | it could not score any candidate, which is indistinguishable from finding none good enough."*

The test still catches the `invalid formula "f": not a call` regression it was written for, because that would error. But it does not demonstrate what its name claims, and a mutation that broke candidate scoring outright would sail straight through.

## Suggested fix

Assert the screen actually worked before asserting what it did not do:

```r
expect_identical(sum(sw$criteria$uncomputable_reasons), 0L)
```

plus an assertion that at least one candidate was scored, or use a base fit whose nuisance block inverts.

Issue #211, opened 2026-09-02
Phase-identifiability warning misdiagnoses single-distinct-time data and names parameters that do not exist


ehrlinger:
Found by the `RELEASE.md` step-5 adversarial review of `git diff v1.2.2...main`, verified at `602e294`.

## What happens

`.hzr_check_phase_identifiability()` (`R/likelihood-multiphase.R:1189`) measures each phase by `variation`, the relative range of its contribution across the observed times. With **one distinct observation time** that is `0` for every phase by construction, being `(max - min) / max` over a single value.

The warning then fires on all phases and misdescribes what happened. With `time = rep(2, 20)` and phases `early = cdf`, `const = constant`:

> Phases 'early', 'const' have a contribution that is constant across the observed times (relative variation 0, 0). The phase has already finished before the first observation, so it acts as a constant offset: 'mu' remains identified but the shape parameters do not... A 'cdf' phase whose half-life is far shorter than the first observed time is the usual cause.

Three claims in that text are wrong for this input:

1. the `constant` phase **has no shape parameters**;
2. neither phase "finished before the first observation";
3. the stated cause is not what occurred.

A single-row fit produces the same message.

## Why it matters

The real condition, one distinct observation time, is a *worse* identifiability problem than the one being reported, and it goes unnamed. A diagnostic written specifically to prevent a misreading is confidently producing one.

## Suggested fix

Guard on `length(unique(time[ok])) < 2` before the `variation` scan and report that case separately, in its own words.

PR #217, opened 2026-09-02
fix(objective): record the objective on the fit so refits keep the estimand

### LINKED ISSUE #197: Ridge detector: fit$weak = NULL conflates "not checked" with "checked, clean"
Deferred from the review of #194, which fixed the three high-severity findings (the missed ridge,
the unnamed parameters, and the vacuous negative test). These are the medium/low remainder.

## 1. `NULL` means four different things, and one of them is a `Suggests` gap

`R/hazard_api.R` and `man/hazard.Rd` both say `weak = NULL` means the fit is well identified.
It is in fact returned for all of:

- the fit was never run (`fit = FALSE`, or `fit = TRUE` without `theta`);
- **no Hessian was obtainable, so `rcond` is `NA`** — precisely the path taken when `numDeriv` is
  absent, since `R/optimizer.R` returns `list(vcov = NA, rcond = NA_real_, pd = NA)`;
- a kept row of `vcov` carries a non-finite entry;
- there genuinely is no ridge.

`numDeriv` is a `Suggests` in this package and the analytic Hessian declines for left- and
interval-censored rows by design, so the second case is reachable on a real install. The
SE-unavailability itself does warn, so that contract is met — but `is.null(fit$weak)` is offered
in `NEWS.md` as the *programmatic* check ("so it can be checked programmatically rather than
scraped from a warning"), and on such an install it reports "well identified" for a fit that was
never examined. A sentinel distinguishing "not checked" from "checked, clean" would fix it.

The same applies to `R/read-outhaz.R`, which assembles a `fit` list outside `hazard()` and so
never has a `weak` element at all — an imported SAS fit reads as well identified.

## 2. Only one direction is ever reported

`.hzr_weak_direction()` returns a single eigenvector's worth of parameters. Two independent
ridges produce one warning naming one pair, and the message's phrasing invites reading the
unnamed parameters as identified. With exactly tied eigenvalues LAPACK may return any vector in
the degenerate eigenspace, in which case the loadings spread and the reported set can be the
union of both ridges described as one.

## 3. Test assertions that restate a guard

In `tests/testthat/test-weak-direction.R`:

- `expect_gt(abs(w$correlation), 0.99)` cannot fail — the function returns `NULL` unless
  `abs(r_max) >= cor_tol` (0.99), and the preceding assertion already establishes non-`NULL`. The
  fixture value is 0.99999; assert that.
- `expect_length(w$params, 2)` is the only test of the unnamed-parameter fallback and asserts
  nothing about the labels. It passes for `c("par1", "par2")`, for `c("", "")` and for
  `c(NA, NA)`.
- `weights` is a documented field of the returned object and is asserted nowhere. Nothing pins
  `share = 0.9`, and nothing would catch the loadings being left un-squared.
- One comment misattributes its own mechanism: `diag(c(0, 1))` returns `NULL` at the zero-variance
  `keep` filter, never reaching the correlation step it claims to test.

## 4. The multiphase fixture sits 0.5% above its own gate

It lands at correlation -0.995 against a `cor_tol` of 0.99, with the ridge eigenvalue 2.0225
against a runner-up of 1.6587. With the same seed and `n_starts = 1` the ridge is not reached at
all, so the house preference for a documented `PARMS` vector with `n_starts = 1` is unavailable
here and the test genuinely depends on the optimizer path. Across BLAS/platform this is a
plausible flake, and it would fail as a *false alarm* — the detector looking broken — rather than
pointing at anything real.

## 5. The bootstrap's select-mode pre-flight does not muffle the new warning

`R/diagnostics.R` muffles by matching `"hzr_safe_solve"` in `deparse(conditionCall(w))`. The ridge
warning is raised with `call. = FALSE`, so `conditionCall(w)` is `NULL` and it is not muffled; it
surfaces once from the up-front screen. Replicates are `suppressWarnings()`-wrapped so there is no
flood. Surfacing once is arguably right, but it is outside that handler's stated design and
should be a deliberate choice rather than an accident of `call. = FALSE`.

## 6. Two enumerations are now incomplete

- `print.summary.hazard`'s roxygen lists the notes the method prints (ill-conditioned, not
  positive-definite, not invertible) and does not mention the ridge note added alongside it.
- `vignettes/inference-diagnostics.qmd` says `summary()` prints a note "when the Hessian is
  ill-conditioned or not positive-definite". Not contradicted, but no longer the whole list.

ehrlinger:
Fixes the most severe finding from the `RELEASE.md` step-5 review of `v1.2.2...main`, plus three documentation defects from the same review, plus the process gap that let them sit in a session transcript.

## 1. `objective` was never recorded on the fit

`hazard(objective = "sas")` was accepted, `match.arg`'d and acted on by the optimizer, and then dropped. `spec` was `list(dist, control, time_windows, phases)`, with no `objective`.

Everything that rebuilds a `hazard()` call from `fit$spec` therefore reverted to `"likelihood"`:

- every `hzr_stepwise()` candidate refit, the Wald-fallback refit, and each accepted move (`R/stepwise-refit.R`);
- the score test's numeric Hessian and gradient (`R/score-test.R:252`, `:489`).

So selecting on a `"sas"` base fit differenced `delta_logLik`, `aic` and `delta_aic` across **two different estimands**, and the score statistic was computed against a likelihood the fit was never fitted to. The two objectives differ by about 22 log-likelihood units on the esophagectomy reference, larger than most single-variable effects. The run produced a full `$steps` table with no error and no warning: the signature defect `AGENTS.md` names.

Worth recording the asymmetry it created: `hzr_bootstrap()` **refit** mode re-evaluates the stored call, which *did* carry `objective = "sas"`, so it was already consistent; **select** mode goes through the scope refit and was not. Nothing marked the difference.

**Fix.** `fit$spec$objective` records it, and `.hzr_fit_objective()` is the single accessor every consumer reads, mirroring the `.hzr_refit_blocker()` idiom so the answers cannot drift. A fit carrying no `objective` predates the argument and is read as `"likelihood"`, which is a fact about those objects rather than a fallback guess.

The single-distribution refit path is deliberately untouched: `hazard()` rejects `objective = "sas"` for anything but multiphase at `R/hazard_api.R:443`, so there is nothing there to preserve.

## 2. Three documentation defects

| Was | Is | Where |
|---|---|---|
| `fit$weak` | `fit$fit$weak` | `NEWS.md` ×4 |
| `fit$phase_share` | `fit$fit$phase_share` | `NEWS.md`, roxygen → `man/hazard.Rd` |
| "rows reading `wald_z` are exactly the ones the fallback rescued" | qualified to `action == "enter"` | `NEWS.md`, roxygen → `man/hzr_stepwise.Rd` |

The first is the sharp one. `is.list(fit$weak)` returns `FALSE` for **every** fit, ridge or not: verbatim the failure the 1.2.8 three-value change was made to prevent, reached by a different route. `man/hazard.Rd` already had it right; `NEWS.md` did not, and `NEWS.md` is what the bullet offered as the programmatic check.

The third was simply false: `R/stepwise.R:291` sets `drop_criterion <- "wald"` under `criterion = "score"`, and `R/candidate-score.R:201` stamps `"wald_z"` at `df == 1` regardless of mode, so with `direction = "both"`, **every drop row** reads `"wald_z"` while `n_wald_fallbacks` may be `0`.

## 3. `RELEASE.md`: file what you do not fix

Step 5's only output is a session transcript, so a finding that is neither fixed nor filed is gone, and the next review pays the full verification cost again. That is not hypothetical: this review returned ten findings and the six not fixed here had no home until they were filed as #211–#216.

The same section now records that re-verification applies to *filing*, not only to fixing, with #215 as the worked example: the review reported `$steps` with zero rows and reason `information_indefinite`; reproducing it gave **one** row and `nuisance_singular`. Conclusion held, mechanism did not, suggested fix keyed on the wrong code.

`RELEASE.md` is `.Rbuildignore`d, so this is documentation-only with no package impact.

## Tests

Five new tests in `tests/testthat/test-sas-interval-objective.R`:

- `objective` is recorded, with **distinct** expected values for both settings, since asserting only the `"sas"` case would pass against an implementation that hardcoded it;
- `.hzr_fit_objective()` returns `"sas"` when present and `"likelihood"` when absent;
- a scope refit keeps the base fit's objective, checked for both settings;
- the two objectives differ numerically by more than 1 log-likelihood unit; this one guards the others from going cosmetic, since if the objectives ever agreed, forwarding the label would protect nothing while the label tests stayed green;
- **the vector-interface × `"sas"` intersection**, added at rebase time (see below).

Mutation-tested throughout. Removing the `$spec` recording fails 6 assertions; removing the refit forwarding fails 1; against the post-rebase code, dropping the objective argument fails 2 and dropping `"time_upper"` fails 1.

## Note on the rebase over #210

#210 and this branch both edit the same `do.call(hazard, ...)` in `R/stepwise-refit.R`. Git conflicted on the **comments** and auto-merged the **argument list**, the half where a dropped argument silently reinstates the bug this branch fixes.

Both halves are present and verified live by the mutation tests above. The rebase also manufactured a combination neither branch owned: a vector-interface multiphase fit under `objective = "sas"`. That is exactly what SAS `SELECTION` jobs land on: 25 of 26 use `ICENSOR`, so they arrive interval-censored through the vector interface, which is why `objective = "sas"` exists, so it now has a regression test.

## Gate

Rebased onto `6a73c1a`.

- `devtools::document()`: clean, no `man/` / `NAMESPACE` drift
- `lintr::lint_package()`: **0 lints**
- `spelling::spell_check_package()`: **clean**
- `test-sas-interval-objective.R`: **0 FAIL / 0 SKIP / 56 PASS**
- Full `devtools::test()` and `R CMD check --as-cran` from a clean `git archive`: **running at the time of opening**; results posted below when they land. Pre-rebase, the same commit was 0 FAIL / 6 SKIP / 3534 PASS with `R CMD check` Status: OK at 199s.

Closes #197 is **not** claimed: that was already closed and verified independently.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,8 +40,11 @@
   `criterion = "score"`, where every neighbouring row carries a Q. The
   selection was right and `p_value` was right, but a reader recomputing a
   p-value from `stat` the way the neighbouring rows permit got an answer wrong
-  by dozens of orders of magnitude. Under `criterion = "score"` the rows
-  reading `"wald_z"` are exactly the ones the fallback rescued.
+  by dozens of orders of magnitude. Under `criterion = "score"` the *entry*
+  rows reading `"wald_z"` are the ones the fallback rescued --- filter on
+  `action == "enter" & stat_type == "wald_z"`. Drop rows are always
+  Wald-tested under that criterion, following SAS, so they read `"wald_z"`
+  whether or not the fallback fired.
 
 * **`hzr_bootstrap()` reports Wald fallbacks in select mode**, through
   `$n_wald_fallback_replicates` and `$n_wald_fallbacks`, with a warning when
@@ -58,6 +61,29 @@
   works, and is what makes the multiple-nomogram warning actionable.
 
 ## Bug fixes
+
+* **`objective` is now recorded on the fit, so refits keep the estimand.**
+  `hazard(objective = "sas")` was accepted and acted on by the optimizer, but
+  the choice was never stored on the object. Everything that rebuilds a
+  `hazard()` call from `fit$spec` therefore reverted to
+  `objective = "likelihood"`: every `hzr_stepwise()` candidate refit, the
+  Wald-fallback refit and each accepted move, plus the score test's numeric
+  Hessian and gradient.
+
+  Selecting on a `"sas"` base fit consequently differenced `delta_logLik`,
+  `aic` and `delta_aic` across *two different estimands*, and the score
+  statistic was computed against a likelihood the fit had not been fitted to.
+  On the esophagectomy reference the two objectives differ by about 22
+  log-likelihood units -- larger than most single-variable effects -- and the
+  run produced a full `$steps` table with no error and no warning.
+
+  `fit$spec$objective` now records it, and one accessor feeds every consumer
+  so they cannot drift apart. A fit that carries no `objective` predates the
+  argument and is read as `"likelihood"`, which is what it was.
+
+  Note the asymmetry this had created: `hzr_bootstrap()` *refit* mode
+  re-evaluates the stored call, which did carry `objective = "sas"`, so it was
+  already consistent; *select* mode goes through the scope refit and was not.
 
 * **A candidate that neither criterion could test is now reported as such.**
   `criterion = "score"` declines a candidate whose observed information is
@@ -86,18 +112,18 @@
   No behaviour changed in what gets selected: a candidate that could not be
   tested is still not entered. What changed is that the run says so.
 
-* **`fit$weak` now distinguishes "no ridge" from "not checked".** The
+* **`fit$fit$weak` now distinguishes "no ridge" from "not checked".** The
   weak-identification diagnostic introduced in 1.2.7 returned `NULL` both when
   a fit had been examined and found well identified and when it could not be
   examined at all -- most importantly when no Hessian was available, which
   happens on an install without the suggested numDeriv package and on fits whose
   rows are left- or interval-censored, where the analytic Hessian declines by
-  design. `NEWS` offered `fit$weak` as the programmatic check, so on those
+  design. `NEWS` offered `fit$fit$weak` as the programmatic check, so on those
   installs it certified as well identified a fit nothing had looked at.
 
   The field now takes three values: a list when a ridge was found, `NULL` when
   the fit was examined and is well identified, and `NA` when the check could
-  not run. Test it with `is.list(fit$weak)` rather than `!is.null()`.
+  not run. Test it with `is.list(fit$fit$weak)` rather than `!is.null()`.
   `summary()` prints a note in the `NA` case saying the check did not run, and
   a fit imported with `hzr_read_outhaz()` -- which has no R Hessian to examine
   -- now reports `NA` rather than reading as certified clean.
@@ -195,6 +221,19 @@
 
 ## Documentation
 
+* **Corrected the documented paths for two fit fields.** The weak-identification
+  result is at `fit$fit$weak` and the phase shares at `fit$fit$phase_share`;
+  `NEWS.md` and, for the shares, `?hazard` had both named them one level too
+  high. Code following the documented recipe got `FALSE` from
+  `is.list(fit$weak)` for every fit, ridge or not -- the same wrong answer the
+  1.2.8 three-value change was made to prevent, reached by a different route.
+
+* **`stat_type` no longer over-claims which rows the Wald fallback rescued.**
+  Under `criterion = "score"` drop rows are always Wald-tested, following SAS,
+  so they read `"wald_z"` whether or not the fallback fired. The rescued rows
+  are the *entry* rows: filter on
+  `action == "enter" & stat_type == "wald_z"`.
+
 * **`predict()` and `hzr_nelson()` now say that SAS draws narrower bands.**
   SAS `%KAPLAN`, `%NELSONT` and `PROC HAZPRED` all take their band width from
   `CLEVEL`, whose default is `0.68268948` -- documented in the macro source as
@@ -262,7 +301,7 @@
   `hazard()` now warns, once per fit, naming the parameters that span the flat
   direction along with their correlation and the Hessian's `rcond`, and
   `summary()` prints the same note. The finding is recorded on the object as
-  `fit$weak`, so it can be checked programmatically rather than scraped from a
+  `fit$fit$weak`, so it can be checked programmatically rather than scraped from a
   warning. See the 1.2.8 notes below for the three values that field takes.
 
   The direction is read off the *correlation* of the estimates rather than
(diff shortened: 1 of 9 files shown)


ehrlinger:
## Gate complete: `d5212d9`, rebased onto `6a73c1a`

Replaces the "running at the time of opening" note in the PR body. All local steps have now finished.

| Step | Result |
|---|---|
| `devtools::document()` | clean, no `man/` / `NAMESPACE` drift |
| `lintr::lint_package()` | **0 lints** |
| `spelling::spell_check_package(use_wordlist = TRUE)` | **clean** |
| `devtools::test()` (`NOT_CRAN=true`) | **0 FAIL / 6 SKIP / 3572 PASS** |
| `R CMD check --as-cran`, with the manual, from a clean `git archive` | **Status: OK**, 0 errors, 0 warnings, 0 notes |

Tarball 2.9 MB; `build/vignette.rds` present; no `CLAUDE` / `AGENTS` / `RELEASE.md` leakage.

The 6 skips are the SAS fixture-availability baseline, unchanged; nothing new became conditional. The 3572 passes reconcile against the 3534 this commit scored before the rebase: #210 brought 30 in via `main`, and the interaction test added 8.

### Mutation results

The tests were checked against deliberately broken code rather than only against working code:

| Mutation | Effect |
|---|---|
| remove `objective` from `$spec` | **6 assertions fail** |
| remove the forwarding from `.hzr_refit_with_scope()` | **1 fails** |
| remove `objective` from the post-rebase `do.call` | **2 fail** |
| remove `"time_upper"` from the carried response fields | **1 fails** |

### ⚠️ Check-time budget is drifting

`AGENTS.md` records 3m29s (209s) at `c9f6c28`, tests 84s, vignettes 42s. Measured here:

| Component | `c9f6c28` | now (`d5212d9`) | Δ |
|---|---|---|---|
| Overall | 209s | **245s** | +36s |
| Tests | 84s | **99s** | +15s |
| Vignette rebuild | 42s | **43s** | +1s |

Still comfortably inside CRAN's ~10-minute ceiling, and this PR is not the main cause; `main` alone measured 212s earlier today. But the direction is consistent across two release cycles now, and it is tests that are moving. Worth re-stamping the `AGENTS.md` table with a dated entry rather than letting the figure go quietly stale.

### 🔴 One open item: Copilot's review is correct

Reproduced at `d5212d9`:

```
ERROR: formal argument "objective" matched by multiple actual arguments
```

`.hzr_refit_with_scope()` now always supplies `objective = .hzr_fit_objective(current)`, but still forwards `...` as `extra_args`, which filters only `weights` and `time_windows`. `hzr_stepwise()` documents `...` as "Passed to the underlying `hazard()` refits", so `hzr_stepwise(fit, objective = "sas")` reaches `do.call()` with the argument twice. **This is a regression introduced by this PR**; before it, the call worked.

The one-line fix (add `"objective"` to that filter) is not obviously the right one: it would *silently discard* an explicit user argument, which is the failure mode this package keeps getting bitten by. Deciding between silently dropping it and refusing it with an explanation, pending maintainer input. **Not ready to merge until that lands.**

Issue #224, opened 2026-09-04
A phase named 'time' silently destroys the prediction-time column of predict(decompose = TRUE)


ehrlinger:
Found by the `r-reviewer` pass over the #214 fix (`4387e39`), and verified on `main` at `2cf7a35`.

## What happens

`R/hazard_api.R:1316-1318` builds the decomposed prediction frame by seeding it with the time column and then writing one column per phase into the same frame:

```r
out <- data.frame(time = pred_time, total = result$total)
for (nm in names(phases)) {
  out[[nm]] <- result[[nm]]
}
```

When `nm == "time"`, the assignment overwrites the prediction times with that phase's cumulative-hazard contribution. The phase gets no column of its own, and the times are gone.

## Reproducer

```r
set.seed(1); n <- 80
dat <- data.frame(t = rexp(n, 0.2) + 0.01, status = rbinom(n, 1, 0.7))
f <- hazard(survival::Surv(t, status) ~ 1, data = dat, dist = "multiphase", fit = TRUE,
            phases = list(time = hzr_phase("cdf"), late = hzr_phase("constant")))
predict(f, type = "cumulative_hazard", decompose = TRUE)
#>   time      total       late      <- 3 columns for a 2-phase model, not 4
#> 1    0 0.46175915 0.46175915
```

The `time` column now holds the `time` phase's contribution; the actual prediction times are not in the frame at all. No error, no warning, a populated data frame — this is verbatim the "output that looks like a result and is not" shape from `AGENTS.md`.

## Relationship to #214

Identical in shape to #214 (a phase name colliding with a reserved key in a container it shares) but a **different key and a different container**: #214 was `total` in the cumulative-hazard accumulator, this is `time` in the decomposed prediction frame. #214's guard reserves only `total`, deliberately, so this is untouched by it.

Note the reserved set is path-dependent: for the wide `predict(decompose = TRUE)` frame it is `c("total", "time")`; for the long-format `predict_cl()` output (`R/predict-cl.R:507`) it is only `total`, because there phase names land in the `component` factor and `time` is a separate column with no clash.

## Suggested fix

Extend the reservation in `.hzr_validate_phases()` to `time` as well, which is one line and cannot drift. The alternative — building the phase columns in a separate frame and `cbind`-ing — leaves the duplicate-column-name problem to `data.frame()` and does not help `predict_cl()`.

[...]

PR #225, opened 2026-09-04
fix(phases): reject 'total' as a phase name (#214)

### LINKED ISSUE #214: A phase named 'total' collides with the contribution accumulator

## What happens

`R/likelihood-multiphase.R:201` writes the accumulator into the same list as the per-phase contributions:

```r
phase_contributions$total <- total
```

A user phase legitimately named `total` is overwritten by it. `.hzr_validate_phases()` does not reject the name.

This is pre-existing (the CoE path already consumed the collision), but `.hzr_phase_shares()` is new code that reads `contrib$total` alongside `contrib[[nm]]`, so `phases = list(total = hzr_phase("cdf"), late = ...)` now *also* reports `share == 1` for that phase, and the identifiability diagnostic reads it as dominating the fit.

## Suggested fix

Reject `total` as a phase name in `.hzr_validate_phases()`, which is cheapest and cannot drift, or key the accumulator on a name that cannot collide.

ehrlinger:
Closes #214.

## The defect

`total` is the key `.hzr_multiphase_cumhaz()` stores its accumulator under, so a phase of that name lost its own contribution vector to it. Reproduced on `main` before the fix:

```
names(contrib): total, late
length: 2  -- 2 phases + accumulator should be 3
  phase     share variation
1 total 1.0000000 0.8722437
```

`.hzr_phase_shares()` then read the accumulator as that phase's contribution and reported `share == 1`, which the identifiability diagnostic at `R/likelihood-multiphase.R:1234` takes as a phase dominating the fit. `hazard()` returned a normal, populated fit throughout — no error, no warning.

## The fix

One guard in `.hzr_validate_phases()` (`R/phase-spec.R:681`), placed after auto-naming so `phase_N` names cannot trip it. All three `R/` callers reach it, and the only other phase-construction site (`R/read-outhaz.R:207`) hardcodes `early`/`constant`/`late`.

Reserving the name rather than renaming the accumulator is the right way round: `total` is *documented public output* — the decompose column at `R/hazard_api.R:959` and the ordered factor level at `R/predict-cl.R:500` — so renaming it would be the breaking change. Any script that used a `total` phase was already getting a wrong answer, so erroring is strictly better than what it had.

The guard also closes two collisions on the same name the issue did not mention: `predict(decompose = TRUE)` labels its summed column `total`, and `predict_cl()` builds its component levels as `c("total", names(phases))`, where the phase name would have produced a duplicated factor level.

## Tests

Four, all verified failing without the guard (run against the parent commit in a throwaway worktree, not assumed):

- the validator directly;
- `hazard()` with `fit = TRUE`, the shape a user actually writes;
- `hzr_theta_names()`, which is exported and validates through the same path — so its behaviour changed too, and NEWS says so;
- a cross-check that reads the accumulator key **off the returned object** rather than restating the guard's error string. Asserting only that `"reserved"` is thrown would stay green if the accumulator were renamed, leaving the guard protecting a name that no longer collides. Mutation-tested: renaming `phase_contributions$total` to `$sum` leaves the three `"reserved"` tests green and fails this one alone.

## Scope

A phase named `time` collides with the prediction-time column of `predict(decompose = TRUE)` in the same shape and with the same silence. That is a different key in a different container, and is filed as #224 rather than widened into this fix.

Pre-existing fits saved by 1.2.8 or earlier still contain the bad phase list and still return the overwritten decomposition; `predict.hazard()` reads `object$fit$phases` without re-validating. This fix is fit-time only.

## Gates

- `devtools::document()` clean, `lintr::lint_package()` **0 lints**, `spelling::spell_check_package()` clean
- `devtools::test()` (`NOT_CRAN=true`): **0 FAIL / 6 SKIP / 3588 PASS** — skips are SAS fixture availability only
- `R CMD check --as-cran` from a clean `git archive` export, with the manual and vignettes: **Status: OK** (tests 80s/87s, vignettes 42s/46s, tarball 2.89 MB, no `CLAUDE`/`AGENTS` files)

Version bumped 1.2.8 → **1.2.9**, patch digit only; `NEWS.md` heading updated by hand to match, since nothing in this repo greps for the mismatch.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: TemporalHazard
 Title: Temporal Parametric Hazard Modeling
-Version: 1.2.8
+Version: 1.2.9
 Authors@R: person("John", "Ehrlinger", email = "john.ehrlinger@gmail.com", role = c("aut", "cre", "cph"))
 URL: https://ehrlinger.github.io/TemporalHazard/, https://github.com/ehrlinger/TemporalHazard
 BugReports: https://github.com/ehrlinger/TemporalHazard/issues
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,18 @@
+# TemporalHazard 1.2.9
+
+## Bug fixes
+
+* **A phase named `total` is now rejected rather than silently losing its
+  contribution.** `total` is the key the multiphase cumulative-hazard
+  accumulator is stored under, so `phases = list(total = hzr_phase("cdf"), ...)`
+  overwrote that phase's own contribution vector with the summed hazard. The fit
+  returned normally; the phase then reported a contribution share of exactly 1
+  and the identifiability diagnostic read it as dominating the model. The name is
+  reserved across the decomposition and prediction output as well, where it
+  labels the total column. `hazard()` and `hzr_theta_names()` -- which applies
+  the same validation -- now stop with an explanatory message, and the
+  reservation is documented on the `phases` argument (#214).
+
 # TemporalHazard 1.2.8
 
 ## New features
--- a/R/hazard_api.R
+++ b/R/hazard_api.R
@@ -161,7 +161,10 @@ NULL
 #'   and requires `phases`.  See the **Baseline distributions** section for what
 #'   each means and when to use it.
 #' @param phases Optional named list of [hzr_phase()] objects specifying the
-#'   phases for a multiphase model (`dist = "multiphase"`).  See Examples.
+#'   phases for a multiphase model (`dist = "multiphase"`).  Names must be
+#'   unique, and `"total"` is reserved: it labels the summed column in the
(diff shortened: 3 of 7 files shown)


ehrlinger:
## Codex review handoff

No actionable findings.

Reviewed the exact live range `2cf7a35…4387e39`. The centralized validation correctly covers `hazard()` and `hzr_theta_names()`, and the regression tests verify both rejection and the underlying accumulator-key invariant.

Verification performed:

- `devtools::document()`: clean
- `lintr::lint_package()`: 0 findings
- Full `devtools::test()`: 3,588 passed, 6 fixture-related skips, 0 failures
- Focused phase tests: 118 passed, 0 failures
- Working tree remained clean after documentation generation
- All GitHub CI checks were green across the five R CMD check platforms

Residual scope is explicit and non-blocking: existing serialized malformed fits are not repaired, and the separate `time` collision remains tracked independently.

Issue #226, opened 2026-09-04
Vector interface does not translate Surv() censoring codes, so left- and interval-censored rows are silently refitted as right-censored


ehrlinger:
Found by the `r-reviewer` pass over the #213 fix, and verified at `2cf7a35` (pre-existing, not introduced by that change).

## What happens

`AGENTS.md` states the rule: this package codes censoring `-1` left, `0` right, `1` event, `2` interval, while `survival::Surv()` under `type = "interval"` uses `0`/`1`/`2`/`3` for right/event/left/interval. It also says the formula path translates in `.hzr_parse_formula()` and **the vector path does not**.

`R/hazard_api.R:556-557` unclasses a `Surv` status column without translating it, so those codes reach the likelihood unchanged.

## Reproducer

```r
ph <- list(early = hzr_phase("cdf"), late = hzr_phase("constant"))
s  <- survival::Surv(c(1,2,3,4,5,6), c(1,0,1,1,0,1), type = "left")

f <- hazard(time = c(1,2,3,4,5,6), status = s, dist = "multiphase",
            phases = ph, fit = FALSE, objective = "sas")
f$data$status
#> 1 0 1 1 0 1        <- rows 2 and 5 are LEFT-censored; stored as right-censored
```

The same `Surv()` through the formula interface stops correctly:

```r
d <- data.frame(tt = c(1,2,3,4,5,6), ev = c(1,0,1,1,0,1))
hazard(survival::Surv(tt, ev, type = "left") ~ 1, data = d, dist = "multiphase",
       phases = ph, fit = FALSE, objective = "sas")
#> Error: objective = "sas" does not support left-censored rows (status == -1) ...
```

No error, no warning, a fit over the wrong censoring model. This is the defect `AGENTS.md` records the package having already shipped once (`Surv(type = "left")` read left-censored rows as right-censored).

`type = "interval"` has the same shape with an extra wrinkle: status `3` (interval) is matched by no branch of `.hzr_logl_multiphase()`, so those rows contribute nothing to the likelihood at all.

## Consequence for #213

The entry guard added in #213 (`.hzr_check_sas_data()`) checks the codes it is given, so on the vector path a `Surv()` left-censored row is invisible to it. Its `@details` and the `NEWS.md` entry both say so explicitly rather than implying coverage it does not have. Closing this issue would let that caveat be removed.

## Suggested fix

[...]

PR #227, opened 2026-09-04
fix(sas): check the SAS objective data preconditions at entry (#213)

### LINKED ISSUE #213: SAS-objective data guards surface as optimizer failures, inviting a remedy that cannot work

## What happens

`inst/dev/SAS-INTERVAL-OBJECTIVE-DESIGN.md` §3 says a data defect "must not be survivable" and specifies `stop()`. `.hzr_check_sas_status()` (`R/likelihood-multiphase.R:469`) does stop, but it is called from **inside** the objective, at `:597` (log-likelihood) and `:712` (gradient). `.hzr_optim_multiphase()`'s per-start `tryCatch` therefore catches it and reframes it.

With `status = c(1, 0, -1, 1, 0, 1)` and `objective = "sas"`:

> Multiphase optimization produced no usable fit from 1 start: 1 errored, 0 ended where the likelihood is not defined... Error(s) raised: objective = "sas" does not support left-censored rows...

The underlying text survives, but the framing says *convergence problem*, which invites the user to raise `n_starts`, a remedy that cannot possibly work, because the condition is a pure function of the data and identical at every start.

The `upper > lower` guard has the same shape.

## Suggested fix

Check both in `hazard()` once, beside the existing `dist != "multiphase"` guard at `R/hazard_api.R:443`, so a data defect is reported as one before any optimization starts.

ehrlinger:
Closes #213.

## The defect

`objective = "sas"` imposes two conditions on the data — no left-censored rows (`PROC HAZARD` has no left-censoring statement) and a positive width on every interval row (the interval-mean hazard divides by `u - l`). Both were checked *inside* the objective, so `.hzr_optim_multiphase()`'s per-start `tryCatch` caught them:

```
Multiphase optimization produced no usable fit from 5 starts: 5 errored,
0 ended where the likelihood is not defined... Error(s) raised: objective =
"sas" does not support left-censored rows...
```

The text survives; the framing says *convergence problem*, which invites raising `n_starts` — a remedy that cannot work, because the condition is a pure function of the data and identical at every start.

Now:

```
objective = "sas" does not support left-censored rows (status == -1): PROC
HAZARD has no left-censoring statement, so no SAS run corresponds to the result.
```

## The fix

`.hzr_check_sas_data()`, called once from `hazard()` before any optimization. Two details:

- It normalises the bounds exactly as `.hzr_logl_multiphase()` does, so it cannot disagree with the objective about which rows offend.
- It reports the offending row **in the data** (`3`) rather than its position among the interval rows (`1`), which is the number a user can actually look up. The inner guard only ever sees `lower[idx_interval]`.

**The guards inside the objective and gradient are deliberately kept.** The gradient is reachable without `hazard()` — `.hzr_score_test()` calls it directly — so entry validation is not guaranteed to have run. Mutation-tested both ways: removing the entry guard fails 5 tests, neutering the inner guard fails 5 *different* ones, so neither layer is decorative.

## Behaviour change

Per the maintainer's decision, the guards fire regardless of `fit`. `hazard(..., fit = FALSE, objective = "sas")` on left-censored data previously returned a normal object carrying an objective it could never be fitted with. Stated in `NEWS.md` and on `@param objective`, not left implied.

## What review changed

The first commit's parity claim was unpinned — **two mutations survived all 70 assertions**: giving up whenever a bound was `NULL`, and normalising `lower` from `time` instead of `time_lower`. Both new fixtures passed the bounds explicitly *and* set them equal to `time`, so neither could tell the implementations apart. The `NULL`-bounds case is the live one: interval rows with no bounds default both to `time`, giving zero width — the commonest way to reach the guard, untested.

Added a no-bounds case, a case where `time` and the bounds disagree in both directions, and a direct assertion that the entry check and the objective reach the same verdict on the same input. Both mutations now fail (8 and 1). The "accepts valid data" control was also moved to `fit = TRUE` so the objective is actually evaluated on data the guard declared valid.

Also from review: an `NA` in `status` made `any(status == -1)` NA-poisoned, throwing a bare `missing value where TRUE/FALSE needed` — pre-existing, but newly reachable at `fit = FALSE` because of this change, so it is named properly here.

## Known limit, documented not hidden

This guards the codes `hazard()` is given. On the **vector** interface a `Surv()` object is unclassed without translating its codes, so a left-censored row arrives as `1` and is invisible to the check; the formula path translates and is guarded. Pre-existing defect of the vector path, stated in the `@details` and `NEWS.md` and filed as #226 rather than widened into this fix.

## Gates

- `devtools::test()` (`NOT_CRAN=true`): **0 FAIL / 6 SKIP / 3607 PASS**
- `R CMD check --as-cran` from a clean `git archive` export, with manual and vignettes: **Status: OK**

Version 1.2.8 → **1.2.9**, patch only. Note this **collides with #225**, which takes the same version from the same base — deliberately, since two patch fixes in one cycle should not inflate to 1.2.10. Whichever merges second needs a trivial `DESCRIPTION`/`NEWS.md` rebase; there is no code conflict.


ehrlinger (commented):
[P2] The new parity test never reaches the objective it claims to compare. In `tests/testthat/test-sas-interval-objective.R:661-666`, `hazard()` calls `.hzr_check_sas_data()` before optimization, so every failing fixture exits through the same entry guard used to compute `entry`; a divergence in `.hzr_logl_multiphase()` / `.hzr_logl_interval()` would still leave `entry` and `inner` identical. Please invoke the internal objective directly with the same normalized-bound fixtures (the existing `sas_mp_*` helpers appear sufficient), so this assertion genuinely pins the duplicated validation.

The implementation path otherwise looks sound, and the targeted test file passes locally (85 assertions). GitHub currently reports the PR as conflicting with `main`, and no CI check runs are attached yet, so those integration results remain outstanding.

Issue #228, opened 2026-09-04
The saturated identifiability message can be false for interval-censored and left-truncated fits


ehrlinger:
Found by agent review of the #211 fix (branch `fix/identifiability-tied-times`). **Pre-existing on `main`**, not introduced by that branch.

## What happens

`.hzr_check_phase_identifiability()` measures `variation` over `time` only. For an interval-censored or left-truncated fit the likelihood also evaluates `Λ` at `time_lower` and `time_upper`, so a phase can be flat across `time` while its shape is identified through those other points.

The message asserts otherwise:

> 'mu' remains identified but the shape parameters do not, and the likelihood is unchanged whether they are pinned or fitted.

## Reproducer

An interval-censored fit whose **lower** bounds sit in the phase's climbing region:

```r
# t_half = 0.1, time = seq(5, 10), time_lower = seq(0.001, 0.05), status = 2
# emitted: "... relative variation 0 ... the likelihood is unchanged whether
#           they are pinned or fitted."
# measured: logLik(t_half = 0.1) = -53.45  vs  logLik(t_half = 0.01) = -58.33
```

**4.88 log-likelihood units** between two values of a parameter the message says the likelihood is unchanged by. The same shape occurs for a left-truncated fit whose entry times lie in the climbing region.

## Relationship to #211

#211 is about this diagnostic naming a cause that did not occur, and the fix for it added a `other_times` argument so the *degenerate-times* verdict is withheld when entry times or interval bounds add an evaluation point. The **per-phase saturated** message is deliberately not withheld there, because withholding it meant that adding a `start` column silently removed a correct diagnostic — a check that does nothing, which is indistinguishable from a clean pass.

So the two failure modes trade against each other, and closing this one properly means measuring `variation` over the points the likelihood actually evaluates rather than suppressing the message. That is a larger change than #211's scope, which is why it is filed rather than folded in.

## Suggested fix

Compute each phase's contribution at the union of the evaluation points (`time`, and the finite `time_lower`/`time_upper` where the status makes them live) and take `variation` over that. Note the alignment constraint: with covariates, `mu_j` varies by row, so the phase contributions cannot simply be concatenated across a longer time vector — the per-row structure has to be preserved, which is why `.hzr_phase_shares()` measures over `time` alone today.

Until then, the limitation is documented on `phase_share_tol` in `?hazard` and in `NEWS.md` rather than left implicit.

PR #229, opened 2026-09-05
fix(multiphase): report degenerate observation times as their own condition (#211)

### LINKED ISSUE #211: Phase-identifiability warning misdiagnoses single-distinct-time data and names parameters that do not exist

ehrlinger:
Closes #211.

## The defect

`variation` is the relative range of a phase's contribution across the observed times, so it collapses for **every** phase when those times carry nothing that separates them. The saturated scan fired anyway and named a cause that had not occurred. With `time = rep(2, 20)`:

> Phases 'early', 'const' have a contribution that is constant across the observed times ... **the shape parameters** do not [remain identified] ... A 'cdf' phase whose **half-life** is far shorter than the first observed time is the usual cause.

A `constant` phase has no shape parameters and no half-life, neither phase had finished before the first observation, and a single-row fit produced the same text.

## The fix

The condition is now asked of the **times**, and the threshold is derived rather than chosen.

With the event times tied, each row's contribution depends on the parameters through `Λ(T) − Λ(e)` at each distinct entry `e`, plus `h(T)`. So `k` added evaluation points supply `k + 1` functionals, and locally identifying `p` **free** parameters needs `k + 1 ≥ p`. Below that a flat manifold passes through every interior point.

Three things follow, and each was a separate defect on the way here:

- **Both halves are required.** The times' relative range must be degenerate *and* no measurable phase may still vary. The range alone condemns times bunched far from the origin, where a steep phase is fully identified — a `g3` case moved the log-likelihood ~11 units while the message called it unidentified. The phase measures alone condemn well-spread times whenever one phase saturates; with a single phase, "every phase is flat" *is* "this phase saturated".
- **Points that add nothing don't count.** `Λ(0) = 0`, and a bound equal to an event time restates it. `.hzr_parse_formula()` synthesises both for every row under `Surv(type = "interval")` — counting them silenced this warning on #211's own input written that way.
- **The count is of free parameters.** Pinning a phase's shapes lowers the bar correctly, which `length(theta)` would not.

## Why the counting rule, and not a threshold

Two thresholds preceded it and both were wrong in the silent direction: `> 0L` silenced a constant entry time (two functionals against five parameters), and `>= 2L` was off by a factor of two for this package's default two-phase model — two distinct entry times cleared it on a fit with an *exact* flat direction (`t_half` 0.5→0.35 with `m` −0.4→−0.731217 matches all three functionals to residual zero).

The evidence that first justified `> 0L` was itself wrong: "moving `t_half` moves the log-likelihood 8 units" was measured with `mu` **held fixed**, which tests sensitivity, not identification. Constructing the flat direction directly — solving the 2×2 system for the two `mu` at fixed shape — holds both functionals constant while `t_half` moves by a factor of five, both `mu` strictly interior, `logL` unchanged to 1e-12. A profile suggests otherwise only because the optimiser collapses one `mu` to ~1e-16 and lands on a cone edge.

The derivation also removed a special case rather than adding one: a lone `constant` phase has `p = 1`, so the single functional tied times supply determines its `mu` exactly (matches the closed form `D/(nT)` to 1e-5) and nothing is warned about. An intermediate version warned "the phase shapes are not identified" there — #211 verbatim, reproduced by its own fix. Two `constant` phases still warn, correctly, but the wording says the phases cannot be told apart rather than naming shapes they do not have.

## Behaviour, verified end to end

| Input | Verdict |
|---|---|
| `time = rep(2, 20)`, two-phase | degenerate-times message |
| Near-tie `c(100, 100.000001)` | same (the measure is relative) |
| Same data as `Surv(type = "interval")` | same (synthesised bounds don't count) |
| Bunched-far times, steep `g3` still varying | silent |
| Well-spread times, one phase saturated | ordinary saturated message |
| Absent phase, well-spread times | absent message, unmasked |
| Constant entry time, tied exits | degenerate-times message |
| 200 distinct entry times | silent |
| Lone `constant` phase, tied times | silent |
| `tol = 0` | silent on every path |

## Known limitation

The per-phase measures are taken over the event times alone, and this is a *local* identification argument. That cuts both ways: a phase can be flat across the event times while identified through the bounds (the saturated message can overstate), and a fit can clear the counting rule and still have a flat direction the guard cannot see (it can under-warn, which is the half that fails silently). Both directions are now stated in `NEWS.md` and `?hazard` and tracked in #228, rather than left implicit.

## Process

This took six designs and six adversarial review rounds, each of which found a real defect — including two where the fix reintroduced the defect class #211 is about, and one pair (rounds 4 and 5) that contradicted each other and had to be adjudicated by constructing the flat direction explicitly. The recurring cause was choosing a predicate that *correlates* with the condition instead of one that computes it. Details are in the commit messages, which are kept unsquashed for that reason.

Two tests of mine were themselves defending bugs: one asserted a warning for data whose shape is unidentified only under a broken metric, and one compared a computation with itself (`Surv` passed as `status` on the vector path, where `unclass(status)[, 2L]` reads the matrix positionally, so both loop iterations ran the same fit).

## Gates

- `devtools::test()` (`NOT_CRAN=true`): **0 FAIL / 6 SKIP / 3682 PASS**
- `R CMD check --as-cran` from a clean `git archive` export, with manual and vignettes: **Status: OK**, tests 82s/87s — below `main`'s pre-branch 91s, after marking five end-to-end fits `skip_on_cran()` (CI sets `NOT_CRAN=true`, so all five still run on all five platforms)
- Six mutations designed against the final rule, all killed

No version bump: `main` is already accumulating 1.2.9, so this is a third bullet under that heading.


ehrlinger (commented):
BLOCKING REVIEW

I found two issues that need resolution before merge: the production call site can suppress the new warning using bound values that do not enter the likelihood, and the newly added end-to-end test fails the required Windows check. GitHub does not permit an author to request changes on their own PR, so this is submitted as COMMENT but is blocking.

Verification: reviewed the immutable diff from 6e68f8041d8e3538adc0aaace14ffe62e75391b9 to 4fb2d81910fadbab9506986ee57115d8c62ccf90; reproduced the production false negative in a clean git-archive snapshot; the targeted test file passes locally; Windows CI fails at the inline test below.

Non-blocking documentation nit: `hazard()` currently says “add an enough evaluation points”; please change that to “add enough evaluation points.”

R/likelihood-multiphase.R
```diff
@@ -2103,9 +2219,12 @@
   # Identifiability is checked on the fit that won, not on the starts: a phase
   # can be identified at a start and dead at the optimum, and it is the
   # reported estimates a reader will use.
+  # time_lower/time_upper are further points the likelihood evaluates, so a
+  # fit whose `time` is tied can still identify the shapes through them.
   best_result$phase_share <- .hzr_check_phase_identifiability(
     best_result$par, time, phases, covariate_counts, x_list,
-    tol = phase_share_tol)
+    tol = phase_share_tol,
+    other_times = c(time_lower, time_upper))
```
ehrlinger: [P1] Count only bounds used by each row’s likelihood

This concatenates every `time_lower` and `time_upper` value, but `time_upper` affects only left/interval-censored rows (and `time_lower` has status-dependent roles). Consequently irrelevant bounds can silence the warning. At this exact head, with 20 tied right-censored observations, `.hzr_logl_multiphase()` is bit-identical for `time_upper = NULL` and `time_upper = seq(0.2, 1.8, length.out = 20)` (`-2.3322849201275595` in both cases), yet `.hzr_check_phase_identifiability()` emits one degenerate-times warning for the former and none for the latter because those unused upper bounds make `n_added` large. This is the silent failure mode the guard is intended to prevent. Please construct `other_times` from status-gated bounds that actually contribute to the objective (and add a mixed/vector-interface regression test).

tests/testthat/test-phase-identifiability.R
```diff
@@ -172,3 +172,476 @@ test_that("unmeasurable shares warn about nothing", {
...
+test_that("interval bounds count as other times at the call site", {
+  skip_on_cran()  # end-to-end fit; CI runs it, --as-cran need not
+  # The call site passes c(time_lower, time_upper); a version passing only
+  # time_lower is undetectable without an interval-censored case.
+  skip_if_not_installed("survival")
+  n <- 40
+  d <- data.frame(lo = rep(2, n), up = seq(2.1, 6, length.out = n),
+                  st = rep(2, n))
+  w <- character(0)
+  withCallingHandlers(
```
ehrlinger, line 627: [P1] Make this required-check test platform-stable

The required Windows `R-CMD-check (windows-latest, release)` job fails here: this fit produces `Multiphase optimization produced no usable fit from 1 start ... t_half must be a positive scalar` (`[ FAIL 1 | ... | PASS 3264 ]`). The same test passes locally and on the other CI platforms, so the assertion is currently optimizer/platform-dependent rather than a stable check of argument plumbing. Please supply a stable interior `theta`/fixed shape or test the call-site propagation without relying on this unconstrained default fit, then rerun the Windows matrix.

R/likelihood-multiphase.R
ehrlinger: Confirmed and fixed in `092a07a`.

Verified the reproducer before changing anything: with 20 tied right-censored observations, `.hzr_logl_multiphase()` returns `-2.3322849201275595` both with `time_upper = NULL` and with `time_upper = seq(0.2, 1.8, length.out = 20)`, while the guard warned for the first and not the second. A bound the objective never reads cannot contribute to identification, so counting it was the silent failure this guard exists to prevent.

Each bound is now gated by the status that makes it live, traced through `.hzr_logl_multiphase()` rather than assumed: `time_lower` for status 0/1 (counting-process entry) and 2 (interval lower bound), `time_upper` for -1 and 2.

```r
other_times = c(time_lower[status %in% c(0, 1, 2)],
                time_upper[status %in% c(-1, 2)])
```

The regression test asserts both halves -- that the objective is bit-identical with and without the unused bound (`expect_identical`), and that the guard now is too. Mutation-tested four ways, all killed: reverting to the ungated concatenation, dropping either bound entirely, and widening the upper gate to every status.

tests/testthat/test-phase-identifiability.R
ehrlinger: Confirmed and fixed in `092a07a`. You're right that it was asserting optimiser trajectory rather than argument plumbing: it used the bare phase defaults with no `theta`, so where the unconstrained fit landed decided the result -- fine on four platforms, `t_half must be a positive scalar` on the fifth.

The shapes are now pinned (`fixed = c("t_half", "nu", "m")`) with `theta` supplied, so the fit is deterministic. It still tests what it claims: dropping `time_upper` at the call site flips the assertion.

Worth recording why this reached CI and not me -- the test carries `skip_on_cran()`, so a local `--as-cran` run never executes it, while the runners set `NOT_CRAN=true` and do. Local green was never evidence here.

Windows is re-running now; I'll confirm when it reports rather than assume.

ehrlinger:
Both blocking findings are addressed in `092a07a`, with detail on each thread. Summary:

- **Unused bounds could silence the guard.** `time_upper` is read only for left-censored and interval rows, so on tied right-censored data it changed no likelihood value yet suppressed the warning. Each bound is now gated by the status that makes it live.
- **Windows.** The call-site test used the bare phase defaults with no `theta`, so it asserted where an unconstrained optimiser landed rather than the argument plumbing. Shapes pinned and `theta` supplied; it still flips if `time_upper` is dropped at the call site.
- **Nit:** `?hazard` now reads "add enough evaluation points".

Local gates on `092a07a`: `lintr` 0 lints, `spelling` clean, `devtools::test()` 0 FAIL / 6 SKIP / 3685 PASS, `R CMD check --as-cran` from a clean `git archive` export with the manual: **Status: OK** (tests 94s/101s).

**Not yet confirmed:** the CI matrix is still running on this head, including the Windows job that failed before. That job is the point of the second fix, so I am not calling it resolved until it reports.

ehrlinger (commented):
No findings on the updated head `092a07a`; the two blockers from my review of `4fb2d81` are resolved.

The production call now filters `time_lower` and `time_upper` by the censoring statuses for which the likelihood actually reads them, so irrelevant upper bounds no longer suppress the identifiability warning. The end-to-end interval-bound test is now deterministic with pinned shapes and an explicit `theta`, and the required Windows check passes. The documentation wording noted previously is also corrected.

Verification: the targeted phase-identifiability tests pass locally (98 assertions; the 18 reported warnings are expected conditions exercised by the test file), `git diff --check` passes, and all 13 GitHub checks are green. Submitted as COMMENT because GitHub does not permit self-approval.

Issue #184, opened 2026-08-26
.hzr_parse_sas_nomogram() silently returns NULL when the counter column prints as OBS


ehrlinger:
## Summary

`.hzr_parse_sas_nomogram()` drops PROC PRINT's observation-counter column **by name**:

https://github.com/ehrlinger/temporal_hazard/blob/main/inst/sas-parity/helper-sas-parity.R#L620

```r
cols <- cols[cols != "Obs"]
```

SAS's casing for that label is not stable across the corpus. Listings that print `OBS` keep the counter in `cols`, which makes `length(cols)` one too large; the row-width guard

```r
if (length(toks) < length(cols) + 1L) next
```

then rejects **every** data row, `rows` stays empty, and the function returns `NULL`.

## Why this is worse than it looks

The failure is **silent and indistinguishable from a listing that never printed a nomogram**. Callers cannot tell "no table here" from "table here, not parsed" — both are `NULL`. Any parity sweep that treats a `NULL` as "this job printed no nomogram" will quietly undercount its own coverage, and will report the shortfall as a property of the SAS corpus rather than a defect in this parser.

This is the same failure *class* as #183 (only the first nomogram in a listing is kept, silently), but with a wider blast radius: #183 loses tables 2..n, this loses the table entirely.

## Reproduce

Fully synthetic; no study data.

```r
library(TemporalHazard)
E <- new.env()
source(system.file("sas-parity", "helper-sas-parity.R", package = "TemporalHazard"), local = E)

# Layout A -- counter labelled "Obs". Parses.
lst <- tempfile(fileext = ".lst")
writeLines(c(
  "          Obs     YEARS     _SURVIV    _CLLSURV    _CLUSURV    _HAZARD    _CLLHAZ    _CLUHAZ",
  "",
  "            1    0.0821     0.97106     0.96812     0.97373    0.24456    0.22241    0.26891",
  "            2    0.2500     0.94683     0.94276     0.95062    0.09424    0.08597    0.10332",
  ""), lst)
nrow(E$.hzr_parse_sas_nomogram(lst))
#> [1] 2

# Layout B -- identical table, counter labelled "OBS".
lst2 <- tempfile(fileext = ".lst")
writeLines(c(
  "          OBS    MONTHS     YEARS     _SURVIV    _CLLSURV    _CLUSURV    _HAZARD    _CLLHAZ    _CLUHAZ",
  "",
[...]

PR #230, opened 2026-09-05
fix(sas-parity): let the header decide which column is the counter (#212)

### LINKED ISSUE #212: Nomogram counter detection silently deletes a real measurement column

## What happens

`inst/sas-parity/helper-sas-parity.R:743-748`:

```r
if (w == length(cols) + 1L) {
  m <- m[, -1L, drop = FALSE]            # counter emitted but not in header
} else if (identical(m[, 1L], as.numeric(seq_len(nrow(m))))) {
  m <- m[, -1L, drop = FALSE]            # counter named in the header
  cols <- cols[-1L]
}
```

The `else if` rests on a comment asserting that no measurement column is ever a gapless 1-based integer run: "YEARS goes 0.0821, 0.25, 0.5, 1". That is a property of the corpus currently in hand, not of SAS listings.

A counter-suppressed listing whose time grid is whole years hits it:

```
           YEARS     _SURVIV   _CLLSURV   _CLUSURV   _HAZARD   _CLLHAZ   _CLUHAZ
          1.0000     0.97106    0.96812    0.97373   0.24456   0.22241   0.26891
          2.0000     0.94683 ...
          3.0000     0.91050 ...
```

→ returns `names()` = `SURVIV CLLSURV CLUSURV HAZARD CLLHAZ CLUHAZ`, **zero warnings**. The time key is gone, and because `ncol == length(cols)` afterwards, the shape check that would have caught it passes.

## Why it matters

This is a silent column deletion in a parity helper: the listing parses, the badge is green, and the comparison is against a table missing its time key. `AGENTS.md` already records that these parsers "encode what one study's listings happen to print".

The new test *"a leading measurement column is not mistaken for the counter"* uses `YEARS = 0.0821, 0.25` and cannot reach this branch.

## Suggested fix

Require the run to be integer-valued **and** the header to have an unmatched leading name; or warn when the dropped column carried a header label that is not a known counter spelling.

ehrlinger:
Closes #212.

## The defect

`.hzr_parse_sas_nomogram()` identified `PROC PRINT`'s observation counter structurally, as a leading column running exactly `1..n`. That rested on a comment asserting no measurement column is ever a gapless 1-based integer run — "YEARS goes 0.0821, 0.25, 0.5, 1" — which is a property of the corpus in hand, not of SAS listings.

Reproduced on a counter-suppressed listing with a whole-year grid:

```
whole-year grid    ncol 6   warnings: 0   cols: SURVIV CLLSURV CLUSURV HAZARD CLLHAZ CLUHAZ
fractional grid    ncol 7   warnings: 0   cols: YEARS SURVIV CLLSURV ...
```

`YEARS` deleted, zero warnings. `ncol == length(cols)` afterwards, so the shape guard below could not fire: the listing parsed, and the comparison would have run against a table with no time key.

## The fix

Three cases, in order of what the listing tells us:

1. **Width alone** — the header does not name the column, so a data row one token wider settles it.
2. **The label** — the header names it a counter, so it is one, whatever values it carries.
3. **The `1..n` run** — only as a fallback for an unrecognised name, where the column is **kept** and warned about.

A first version required the run *and* the label. Review showed that was wrong in both directions:

- **The run is not necessary.** Page two of a paginated `PROC PRINT` starts at 41, so an `Obs` counter there failed the `1..n` test and was silently kept as a measurement column — the same paginated-listing family as the life table that was read only to page one.
- **`"_n_"` in the label set was unreachable.** Header tokens have their leading underscore stripped one step earlier, so SAS's `_N_` arrives as `N_` and the entry could never match — while the warning told the reader to add the spelling to `counter_labels`, where it already was. Now `"n_"`, matched post-strip, with `_N_` covered in the label test as SAS prints it.

Labels are matched case-insensitively because SAS's casing is not stable across the corpus (`Obs` in one listing, `OBS` in another, which was #184).

## Why keep-and-warn for the ambiguous case

A `1..n` run under an unknown name is genuinely undecidable from the listing: either a counter SAS has started spelling differently, or a measurement that happens to run `1..n`. Keeping it errs toward an extra column, which is visible in `names()` and in the warning; deleting a real one is not visible at all.

The first version justified this by saying "the caller's shape checks catch" an extra column. **There are no such checks** — the only caller indexes by name and never inspects `ncol`, and the shape guard structurally cannot fire in this branch because keeping the column is what makes the widths agree. The decision was right and the stated reason was false, which is worse than no comment. The code now says the warning is the only defence, because it is.

## Verification

Reviewed against the **real SAS corpus** (194 `.lst` files, 29 with a nomogram header, fixtures present): parser output is byte-identical to `main` in names, row counts and first time value across all 29. `test-sas-parity.R` ran for real rather than skipping — 187 passes.

Six mutations, all killed: removing the label branch or the width arm, trimming the label set, matching case-sensitively, dropping in the ambiguous branch, and restoring the unreachable `"_n_"`. The label loop now iterates every member of the set, including `Row` and `_N_`, which the suite never exercised — the set could previously be trimmed silently.

Two coverage gaps closed while here: the width-based arm had no test at all (removing it broke nothing), and nothing asserted that an unlabelled non-`1..n` column is kept *without* a warning, which keeps the warning from becoming noise.

## Gates

- `devtools::test()` (`NOT_CRAN=true`): **0 FAIL / 6 SKIP / 3634 PASS**
- `R CMD check --as-cran` from a clean `git archive` export, with manual and vignettes: **Status: OK**, tests 80s/85s

No version bump: `main` is already accumulating 1.2.9, so this is a further bullet under that heading. Note #229 also adds a 1.2.9 bullet, so whichever merges second needs a trivial `NEWS.md` rebase — there is no code overlap, the two touch different files entirely.


ehrlinger (commented):
No blocking findings. The parser now uses width when the counter is absent from the header, a case-insensitive known-label match when it is named, and keep-plus-warning for an ambiguous unknown `1..n` column. That avoids silently deleting a legitimate whole-year time key and handles paginated counters that do not start at 1.

One non-blocking test-comment contradiction is noted inline: the test says a `1..n` run remains required for known labels, while both the implementation and the following paginated-counter test deliberately establish that the label alone is decisive.

Verification: the exact-head `test-lst-parser-layouts.R` suite passes (94 assertions), `git diff --check` passes, and all 13 GitHub checks are green. Submitted as COMMENT because GitHub does not permit self-approval.

tests/testthat/test-lst-parser-layouts.R
```diff
@@ -118,10 +118,16 @@ test_that("the nomogram parses when the counter column is labelled OBS", {
...
 })
 
-test_that("the counter is found structurally, whatever SAS calls it", {
-  # The fix must not trade one hardcoded label for two. Any label works
-  # because the counter is identified as a leading 1..n run, not by name.
-  for (label in c("Obs", "OBS", "Observation", "#")) {
+test_that("the counter is dropped under any of its known labels", {
+  # The fix for #184 must not trade one hardcoded label for two, so the set is
+  # case-insensitive and holds every spelling the corpus has shown. The 1..n
+  # run is still required -- the label alone does not drop a column -- but it
```
ehrlinger, line 124: This explanation contradicts both the implementation and the next regression test. Known labels are deliberately decisive regardless of their values (for example, a paginated `Obs` column beginning at 41), so a `1..n` run is **not** still required. Please update the comment so future maintainers do not restore the exact condition this PR removes.

Issue #232, opened 2026-09-05
.hzr_check_sas_data() reports NA as a row index when a bound is NA


ehrlinger:
Found by the `RELEASE.md` step-5 release review of `git diff v1.2.8...main` (2026-09-05). Low severity, filed rather than fixed at release time.

## What happens

`R/likelihood-multiphase.R`, in the interval-width guard:

```r
bad <- idx_interval[!(upper[idx_interval] > lower[idx_interval])]
```

An `NA` bound makes the comparison `NA`, so `bad` carries `NA` and the error message reads `... at index/indices NA.` — a row number that identifies nothing.

## Reachability

Not reachable through `hazard()` today: bounds are validated finite at the entry guards before this runs. It **is** reachable by direct internal call, and the test suite calls `.hzr_check_sas_data()` directly, so the path is live in practice.

## Suggested fix

Treat an `NA` bound as its own data defect and name it, the same way `anyNA(status)` now is — rather than letting it fall through the width comparison and surface as an unusable index. `which(!(upper > lower) | is.na(upper) | is.na(lower))` would at least report the right rows.

Issue #231, opened 2026-09-05
hazard() returns a populated fit object for zero-row data


ehrlinger:
Found by the `RELEASE.md` step-5 release review of `git diff v1.2.8...main` (2026-09-05). **Pre-existing**, outside that diff, filed because the zero-row edge case is on the release checklist and this is the package's named signature shape.

## What happens

```r
ph <- list(early = hzr_phase("cdf"), constant = hzr_phase("constant"))
hazard(time = numeric(0), status = numeric(0), dist = "multiphase",
       phases = ph, fit = TRUE)
```

returns a populated `hazard` object with `converged` set, warning only about `rcond = 0`. No data, a fit object anyway.

## Why it matters

Verbatim the shape `AGENTS.md` names as this package's signature defect: an output that looks like a result and is not. A caller that checks `is.null()` or `converged` sees a healthy fit. The `rcond = 0` warning is real but reads as a conditioning problem, not as "there were no observations".

## Suggested fix

Reject zero-row input in `hazard()` alongside the other entry guards, before any optimization — it is a pure function of the data, so the same argument applies as in #213. Worth checking the single-distribution paths at the same time; this was only reproduced for `dist = "multiphase"`.

PR #233, opened 2026-09-05
docs(release): correct the release-review findings for 1.2.9

ehrlinger:
`RELEASE.md` step 5 for the 1.2.9 release — the adversarial review of `git diff v1.2.8...main`, 37 commits.

**No interaction defect** between #225, #227 and #229, which was the main risk: all three touch `R/likelihood-multiphase.R` and `R/hazard_api.R`, and their guards now run in sequence in `hazard()`. Traced and confirmed independent.

What it did find, all verified by running before acting:

| | Finding | Disposition |
|---|---|---|
| 1 | **The Surv-code description is false** in `NEWS.md` and the `.hzr_check_sas_data()` roxygen | fixed here |
| 2 | `?hazard` reads "add an enough evaluation points" | fixed here |
| 3 | The `NA`-status guard is gated on `objective = "sas"` | fixed here |
| 4 | A test comment contradicts the code it documents | fixed here |
| 5 | The counting rule is off by one when a row has entry `0` | documented here |
| 6 | `hazard()` returns a populated fit for zero-row data | filed as #231 |
| 7 | `NA` bound reported as `at index/indices NA` | filed as #232 |

### The one that matters

The release documents its own known gap **incorrectly**. Both `NEWS.md` and the roxygen said a left-censored row "arrives as `1` and is invisible to the check". Verified against `survival`:

- `type = "left"` codes left-censored as **`0`** → arrives as this package's *right-censored*, and is **fitted as one**
- `type = "interval"` codes left-censored as **`2`** → arrives as *interval*, and the guard rejects it for zero width
- a genuine interval row is coded **`3`** → matches no branch of `.hzr_logl_multiphase()` and contributes nothing

Three different wrong answers depending on `type`, one of which silently drops rows — not one invisible row. Both places are corrected, since they will be read as the authoritative description of #226, which should be re-scoped accordingly.

### The behaviour change

An `NA` in `status` is a data defect under every objective, but the guard only fired for `objective = "sas"`. Under the default it still produced the framing #213 exists to remove:

```
Multiphase optimization produced no usable fit from 1 start: 0 errored,
1 ended where the likelihood is not defined, ...
```

which invites raising `n_starts` — a remedy that cannot work on a pure function of the data. The check now runs before the objective gate. Mutation-tested: moving it back inside fails 5 tests.

### Gates

- `devtools::test()` (`NOT_CRAN=true`): **0 FAIL / 6 SKIP / 3703 PASS**
- The pre-fix tarball already checked **Status: OK** with the manual (2.91 MB, no `CLAUDE`/`AGENTS`/`.claude`); I will re-run `--as-cran` on the merged result before tagging.

**This blocks the v1.2.9 tag.** Findings 1 and 2 are user-facing text that would otherwise ship in the manual and release notes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -48,7 +48,10 @@
   removes a correct diagnostic (#211).
 
   Known limitation, unchanged by this work and stated rather than implicit: the
-  per-phase measures are taken over the event times alone. That cuts both ways.
+  per-phase measures are taken over the event times alone, and the count of
+  evaluation points is deliberately conservative -- it does not credit the
+  extra functional a zero entry time makes separately observable, so a fit can
+  be warned about while being identified. That cuts both ways.
   A phase can be flat across the event times while its shape is identified
   through the bounds, so the saturated message can overstate what is lost; and
   the counting rule above is a *local* identification argument for tied event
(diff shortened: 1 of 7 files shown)


ehrlinger (commented):
No code-review findings on the current head.

I verified that the corrected Surv-to-internal status documentation matches the conversion code, that genuine Surv interval code 3 maps to the implemented interval path, and that the moved missing-status check now protects every objective before status-dependent conditionals. The targeted `sas-interval-objective` and `lst-parser-layouts` tests pass, as does `git diff --check`; docs-current, lint, pkgdown, spelling, and house-style were green when reviewed. The R CMD check matrix and coverage job were still running at posting time.

I am posting this as COMMENT rather than APPROVE because the authenticated GitHub account is also the PR author.

PR #237, opened 2026-09-09
fix(sas): a PARMS phase is active iff its MU is specified and positive

ehrlinger:
**Stacked on #235.** Base is `fix/sas-parms-weibull-generalized`, not `main`, so the diff stays one commit — retarget to `main` once #235 merges.

## The divergence

`.hzr_parse_parms()` decided whether to emit a phase from whether that phase's *shape* operands appeared in PARMS (`if (length(early))`, `has_muc <- "MUC" %in% names(mu)`, `if (length(late))`). The reference C keys on the MU instead, and I read each of these rather than inferring them:

| Reference | What it establishes |
|---|---|
| `parmprc.c:13,18,19` | MUE/MUC/MUL are registered through `setparmno()` |
| `setparmno.c:11-14` | sets `C->phase[n] = 1` only under `stmtfld(parmno) > ZERO` |
| `setprmf.c` | registers the seven shape operands; contains no reference to `phase` at all |
| `stmtprc.c:101-122` | an inactive phase has its shape operands zeroed and status flags cleared |
| `setstat.c:9-12` | returns before `*phasevar = 1` for an inactive phase — its covariates are never registered |
| `shape.c:19,23,31` | `SETG1()`, the `muC` branch and `SETG3()` all gate on `Common.phase[1..3]` |
| `modterm.c:18-22` | all three phases off → `ERROR 1001: No phase selected`, job refused |

So `PARMS MUE=0.2 THALF=1 NU=1 TAU=2 GAMMA=1.5` was a two-phase model in R against PROC HAZARD's one — and the phantom carried an invented `mu` start of 0.1 at the head of its theta block, shifting every later block against `.hzr_phase_theta_names()`'s positional labels. Nothing appeared in `$untranslated`.

## What changed

A phase is built iff its own MU was specified and **positive**. That also fixes the constant phase, which was not in the original scope: `has_muc` tested *presence*, so `MUC=0` built a phase PROC HAZARD leaves inactive. `.hzr_parms_default_mu_start` is removed — nothing can now build a phase without a real MU to log, and it was emitting `log(0) = -Inf` into theta for `MUE=0`.

Everything a closed gate discards now leaves a row: a non-positive MU; a phase's shape operands, `FIX` tokens and covariates; and the case where nothing activates at all. That last one matters — without it `sas-parse-job.R:604` omits `dist`, the call falls through to `hazard()`'s default distribution, and `hzr_translate_sas()` reports **5/5 coverage for a job SAS will not run**.

The MU gate is necessary but not sufficient: an active MU whose phase has no shape operand is still *recorded* rather than built, because SAS's defaults are not this parser's and `tau = 2*Tmax/3` (`setg3.c:317`) needs `max(time)`. Unchanged from 82210d4.

## Corpus impact: none

No job in the public `hazard` corpus is affected — the only PARMS statement this gate changes is a documentation template with literal `?` placeholders. This closes a latent divergence rather than changing any translation already run.

> I could not reproduce the "38 live PARMS blocks" figure cited in #235's test comments; three counting methods give 16, 22 and 38. The "0 affected" conclusion holds under all three, so NEWS names the single affected item instead of asserting a denominator.

## One test rewritten

A #235 test asserted *"the alpha = 1 guard does not fire on a job with no late phase"* via `nrow(untranslated) == 0` — a proxy that held only because nothing else could produce a row. A stray `FIXALPHA` there is real dropped material (`stmtprc.c:118-121` clears its status flag) and now earns one, so the assertion was rewritten to test its stated intent.

## Gate

- `devtools::document()` — no diff in `man/`, `NAMESPACE`, `DESCRIPTION`
- `lintr::lint_package()` — **0**
- `spelling::spell_check_package()` — 0
- `devtools::test()` (`NOT_CRAN=true`) — **0 FAIL / 6 SKIP / 3766 PASS** (skips are SAS fixture availability only)
- `R CMD check --as-cran` with the manual, from a clean `git archive` — **Status: OK**, 3m 25s, tarball 2.9 MB, tests 84s, vignettes 45s, under-check 2742 passes
- Ten mutants killed across two rounds, including a false-positive guard: firing the `modterm.c` refusal when a MU *is* active breaks 3 tests

## Deliberately out of scope (filed separately)

1. `.hzr_parms_early_default` is `nu = 1, m = 0` (R's `hzr_phase()` defaults) but SAS's are `nu = 2, m = 1` (`stmtprc.c:30-37`). Any *partially* specified phase gets a different starting vector. Zero corpus blocks are partially specified, so latent.
2. `has_phases == FALSE` was already reachable via the no-PARMS path (`sas-parse-job.R:598-603`). Read with `modterm.c`, a job with no PARMS has no active phase either — so that older path may also emit runnable fits for jobs SAS rejects. Needs the C question settled first (is `modterm()` universal or multiphase-only?) before anything changes.

Also worth a maintainer decision: this makes the no-phase-selected case **visible** in `$untranslated` rather than emitting `stop()` as the LCENSOR+ICENSOR path does at `sas-parse-job.R:527-540`. A hard refusal would have to change job-level code that also governs (2).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,32 @@
 
 ## Bug fixes
 
+* **`hzr_translate_sas()` no longer builds a phase that `PROC HAZARD` would
+  not.** A `PARMS` statement names its phases with `MUE`, `MUC` and `MUL`; the
+  shape operands (`THALF`/`NU`/`M` early, `TAU`/`GAMMA`/`ALPHA`/`ETA` late)
+  only shape a phase that already exists. The translator had this the other way
+  round and keyed on the shape operands, so
+  `PARMS MUE=0.2 THALF=1 NU=1 TAU=2 GAMMA=1.5` emitted a two-phase model
+  against `PROC HAZARD`'s one -- carrying an invented `mu` starting value of
+  0.1 for the phase that should not have been there, and offering nothing in
+  `$untranslated` to say so. In the reference C only `setparmno()` sets
+  `C->phase[n]`, and only when that `MU` is greater than zero
+  (`src/hazard/setparmno.c`); the seven shape operands are registered by
+  `setprmf()`, which never touches it. A phase is now built only when its own
+  `MU` was specified and positive -- so a `MUC` of exactly zero no longer
+  builds a constant phase either, where before the translator tested only
+  whether the keyword was present. Shape operands belonging to a phase that
+  never activated are recorded in `$untranslated` rather than dropped, since
+  `PROC HAZARD` zeroes them (`src/hazard/stmtprc.c`) and skips their covariates
+  (`src/hazard/setstat.c`). A `PARMS` that activates no phase at all is now
+  recorded too, because `PROC HAZARD` refuses that job outright
+  (`src/hazard/modterm.c`, `ERROR 1001: No phase selected`) where the
+  translator would previously have emitted a runnable single-distribution fit
+  and reported full coverage. No job in the public `hazard` corpus is affected:
+  the only `PARMS` statement in it that this gate changes is a documentation
+  template with literal `?` placeholders, so this closes a latent divergence
+  rather than changing any translation you have already run.
+
 * **`hzr_translate_sas()` no longer discards the `ALPHA` and `ETA` a `PARMS`
   statement specified alongside `WEIBULL`.** The translator read the bare
   `WEIBULL` keyword as a request to constrain the late phase to
--- a/R/sas-parse-parms.R
+++ b/R/sas-parse-parms.R
@@ -30,12 +30,11 @@
 
 # hzr_phase() default shape values (mirrored here so the theta starting
 # vector agrees with what hzr_phase() itself defaults to when PARMS did not
-# supply a given shape parameter). mu has no hzr_phase() argument -- its
-# only documented default is .hzr_phase_start()'s mu_start = 0.1, used here
-# for a built phase whose PARMS never supplied its MU value.
+# supply a given shape parameter). There is no matching mu default: a phase is
+# built only when its MU was specified and positive, so every built phase has
+# a real MU to log.
 .hzr_parms_early_default <- c(t_half = 1, nu = 1, m = 0)
 .hzr_parms_late_default  <- c(tau = 1, gamma = 1, alpha = 1, eta = 1)
-.hzr_parms_default_mu_start <- 0.1
 
 # FIX<param> keywords are their own grammar tokens (FIXTHALF, not FIX+THALF),
 # and the R parameter name is not always the lowercased SAS spelling --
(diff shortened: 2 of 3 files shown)


ehrlinger:
Note for whoever reviews this: **#238 is stacked on this branch** and resolves the Codex P1 raised here. The maintainer's decision is to merge this one first, so this is context for that, not a request to change anything.

Two things land on `main` with this PR and are fixed by #238 immediately after:

**1. The P1 — operand-only `PARMS` slips the no-phase guard.** The guard at `R/sas-parse-parms.R:483` is keyed on `mu`/`early`/`late`, but `WEIBULL`, `FIXDELTA` and `DELTA=` are consumed without entering any of them:

```
PARMS WEIBULL     has_phases=FALSE  untranslated=0
PARMS FIXDELTA    has_phases=FALSE  untranslated=0
PARMS DELTA=0     has_phases=FALSE  untranslated=0
```

Each emits a runnable default-distribution `hazard()` fit at full token coverage for a job `PROC HAZARD` refuses with `ERROR 1001`. #238 keys the gate on whether any phase is *active* instead, which is what `modterm.c:18` actually tests — that condition never asks what `PARMS` contained — so it covers these three plus the wider no-`PARMS` case in one guard, and emits a `stop()` in place of the fit.

**2. One corpus annotation here is wrong.** `dist/examples/hm.dthar.TGA.sas` is annotated *"PROC HAZARD selects no phase here and refuses the job (modterm.c:18-22 raises ERROR 1001)"*. That file is a documentation template carrying literal `PARMS MUE=? THALF=? NU=? M=1 FIXM MUC=?;`. The reference would reject `?` as a **syntax error**, not via `modterm.c` — so the refusal is attributed to the wrong mechanism. #238 gates the refusal on every operand having been read, which drops that row while keeping all the per-operand ones. It is the single corpus difference between the two branches (1 of 57 translated jobs).

Also worth recording, since it is a citation in this branch's comments that callers grep: `hazard.c:298-301` is off by one. Line 298 is a bare `/* */`; the `1001` test is at 301 and `hzfxit("SEMANTIC")` at 302. `hazard.c:299-302` is the accurate span. Corrected in #238.

ehrlinger:
**Correction to my note above**, caught by a parallel session and verified against the file.

I wrote that `dist/examples/hm.dthar.TGA.sas` "is a documentation template carrying literal `PARMS MUE=? THALF=? NU=? M=1 FIXM MUC=?;`". That describes **one of its two `%HAZARD` blocks**, not the file:

- `%HAZARD(` at line **70** — `PARMS MUE=0.2 THALF=0.08 NU=1 M=1 FIXM MUC=0.001;` at line **74**. Fully valid, activates phases 1 and 2.
- `%HAZARD(` at line **110** — `PARMS MUE=? THALF=? NU=? M=1 FIXM MUC=?;` at line **122**. The post-stepwise final-fit block the header comment tells the reader to fill in.

So anyone checking the claim would grep the file, hit line 74 first, and reasonably conclude I had it wrong.

**The conclusion is unchanged:** the second block still fails in the lexer on `?` and still must not be attributed to `modterm.c`'s `ERROR 1001`. It remains the single corpus difference between this branch and #238. Only the description was imprecise, and it is corrected in #238 (`01a9d08`) in the code comments, the test, and NEWS.

PR #238, opened 2026-09-09
fix(sas): a job with no PARMS is refused by PROC HAZARD, not defaulted

ehrlinger:
Stacked on #237 — **please merge that first**; the base is `fix/sas-parms-phase-keyed-on-mu`, so this diff is two commits.

## The question this had to settle first

#237's MU gate records a `PARMS` statement that activates no phase, but deliberately scoped itself out of the older and wider case: a `PROC HAZARD` job carrying **no `PARMS` statement at all**. Whether that is a defect or correct-as-is turns entirely on one question about the reference C — *is `modterm()` reached for every job, or only once a multiphase model has been selected?*

**It is universal.** Verified against `~/Documents/GitHub/hazard`:

- `modterm()` has exactly one call site, `outmods.c:91`.
- `outmods()` sits in `main`'s straight-line sequence at `hazard.c:296` — no enclosing conditional, no early return between `hzrg()` and it.
- The no-phase test is not a filter that could hide the call; it is one of the three disjuncts at `outmods.c:89` that *fire* it.
- `hazard.c:299-302` routes `errorno = 1001` to `hzfxit("SEMANTIC")`, which exits **before** `results()`.

So the job prints no estimates and never fits.

The other half holds too. Across all of `src/hazard`, `C->phase[]` is written in exactly two places: `stmtprc.c:87` zeroes all three at init, and `setparmno.c:14` sets one to `1` — reachable only from `parmprc.c:13,18,19` (MUE/MUC/MUL) and only under `stmtfld(parmno) > ZERO`. The seven shape operands go through `setprmf()`, which never touches it. (`hzpe.c` also writes `phase[]`, but that is `hazpred`/`libhazp` — a different program.)

"No `PARMS`" and "`PARMS` naming no positive MU" are therefore **one state, not two cases**.

## What was wrong

On `main`, a minimal no-PARMS job translated to:

```r
fit <- hazard(data = D, time = FU, status = .hzr_status, fit = TRUE,
              theta = c(), weights = ifelse(DEAD > 0, DEAD, 1))
```

**0 untranslated rows, 3/3 token coverage** — a runnable single-distribution fit, reported as fully translated, for a job SAS refuses outright. `theta = c()` is the tell: the hollow-object shape, which defeats both `is.null` and a length check on the wrapper.

The corpus test could not catch this. `test-sas-translate-corpus.R:20` asserts `tokens_seen > 0` and `seen >= mapped` — coverage counters, which a wrong-but-runnable fit satisfies cleanly.

## The change

**Refused, not recorded.** `.hzr_parse_parms()` returns `refused`, and `.hzr_parse_job()` emits a `stop()` in place of the `hazard()` call, mirroring the `LCENSOR` + `ICENSOR` path at `sas-parse-job.R:536`. An `$untranslated` row alone leaves the fit chunk in place, so a reader who renders past the callout still gets a converged fit with a populated summary standing in for a job that produced no result.

**Refusal requires comprehension.** It fires only when every operand was understood. `PARMS MUE = 0.2 THALF = 1` (spaces around `=`) parses to nothing here, because `.hzr_parse_hazard()` splits on `" "` — while HAZARD's lexer discards whitespace unconditionally (`hazard_l.l:32`, rule at `:50`), making it a well-formed `parmsopt` (`hazard_y.y:138`) whose job **runs**. Refusing that would stop a job the reference accepts. The per-operand rows still record what was not understood; only the refusal, which is a claim about the reference, requires having understood it all.

**A second `PARMS` accumulates** rather than replacing the first, matching the single field table `parmprc()` reads once after all statements are processed. Overwriting made `PARMS MUE=0.2 THALF=1 NU=1; PARMS FIXNU;` refuse a job that runs.

## Corpus effect

Re-measured A/B across all 110 `.sas` files: **exactly 1 of 57** translated jobs differs from #237, and only by **dropping a false claim**. the **second** `%HAZARD` block of `dist/examples/hm.dthar.TGA.sas` (line 110, its `PARMS` at 122) is a documentation template carrying literal `?` placeholders; #237 annotates it as refused by `ERROR 1001`, but the reference would reject `?` as a **syntax error** long before `modterm.c`. That file's *first* block (line 70, `PARMS` at 74) is a valid `MUE=0.2 THALF=0.08 NU=1 M=1 FIXM MUC=0.001` activating phases 1 and 2 and is unaffected — the template shape belongs to one block, not to the file. Its per-operand rows are unchanged, and **no corpus job is refused**.

An earlier text-matching scan suggested 11 affected jobs; every one was a false positive — matches inside comments, or blocks whose `PARMS` arrives via `%hmmodel;` macro expansion. Matching on the *call head* rather than deparsed text gives the figures above.

## Verification

Tests were mutation-checked rather than reasoned about — restoring #237's narrow scoping and keying the wording on `mu`/`early`/`late` each produce failures. One assertion had to be rewritten because `identical(x[[1]], quote(hazard))` passes unconditionally against the emitted `fit <- hazard(...)`; the helper now unwraps the assignment.

Gates, in AGENTS.md order:

- `devtools::document()` — clean, no `man/`/`NAMESPACE` diff
- `lintr::lint_package()` — **0 lints**
- `devtools::test()` with `NOT_CRAN=true` — **0 failures**, 6 skips, **3789 passes** (skips unchanged from the AGENTS.md baseline)
- `R CMD check --as-cran` from a clean `git archive` export, with the manual — **`Status: OK`**

The `r-reviewer` findings and how each was resolved are in the comment below.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,21 +20,35 @@
   never activated are recorded in `$untranslated` rather than dropped, since
   `PROC HAZARD` zeroes them (`src/hazard/stmtprc.c`) and skips their covariates
   (`src/hazard/setstat.c`). A job that activates no phase at all is now
-  recorded too -- whether its `PARMS` named no positive `MU`, or it carried no
-  `PARMS` statement whatsoever -- because `PROC HAZARD` refuses that job
-  outright (`src/hazard/modterm.c`, `ERROR 1001: No phase selected`) where the
-  translator would previously have emitted a runnable single-distribution fit
-  and reported full coverage. The two are one state rather than two cases:
-  `stmtprc.c` zeroes all three phases at initialization and only `setparmno()`
-  turns one back on, so a job with no `PARMS` has no active phase for the same
-  reason a `PARMS` naming `MUE=0` does. `modterm()` is reached on every job,
-  not only once a multiphase model has been selected -- its one call site in
-  `outmods()` is unconditional in the procedure's main sequence, and
-  `hazard.c` exits on `ERROR 1001` before any results are computed. No job in
-  the public `hazard` corpus is affected:
-  the only `PARMS` statement in it that this gate changes is a documentation
-  template with literal `?` placeholders, so this closes a latent divergence
-  rather than changing any translation you have already run.
+  **refused** rather than translated -- whether its `PARMS` named no positive
+  `MU`, or it carried no `PARMS` statement whatsoever. `PROC HAZARD` does not
+  run such a job: `src/hazard/modterm.c` raises `ERROR 1001: No phase
+  selected` and the procedure exits before computing any results, so there is
+  no fit for a translation to be faithful to. The translator previously
+  emitted a runnable single-distribution fit and reported full token coverage;
+  it now emits a `stop()` in place of the `hazard()` call, as it already did
+  for `LCENSOR` combined with `ICENSOR`. The two cases are one state rather
+  than two: `src/hazard/stmtprc.c` zeroes all three phases at initialization
+  and only `setparmno()` turns one back on, so a job with no `PARMS` has no
+  active phase for the same reason a `PARMS` naming `MUE=0` does. `modterm()`
+  is reached on every job, not only once a multiphase model has been selected
+  -- its one call site in `outmods()` is unconditional in the procedure's main
+  sequence.
+
+  The refusal fires only when every `PARMS` operand was understood. A
+  statement this parser could not read is recorded, operand by operand, but
+  never refused: `PARMS MUE = 0.2 THALF = 1` (spaces around `=`) parses to
+  nothing here while `PROC HAZARD`'s own lexer discards whitespace and runs
+  the job with an active early phase, so refusing it would stop a job the
+  reference accepts. Relatedly, a second `PARMS` statement now adds to the
+  first rather than replacing it, matching the single field table
+  `parmprc()` reads once after all statements are processed.
+
+  One job in the public `hazard` corpus changes, and only to drop a claim that
+  was wrong: `dist/examples/hm.dthar.TGA.sas` is a documentation template
+  carrying literal `?` placeholders, which the reference would reject as a
+  syntax error rather than as `ERROR 1001`. Its per-operand rows are
+  unchanged. No corpus job is refused.
 
 * **`hzr_translate_sas()` no longer discards the `ALPHA` and `ETA` a `PARMS`
   statement specified alongside `WEIBULL`.** The translator read the bare
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,9 +40,9 @@
   never refused: `PARMS MUE = 0.2 THALF = 1` (spaces around `=`) parses to
   nothing here while `PROC HAZARD`'s own lexer discards whitespace and runs
   the job with an active early phase, so refusing it would stop a job the
-  reference accepts. Relatedly, a second `PARMS` statement now adds to the
-  first rather than replacing it, matching the single field table
-  `parmprc()` reads once after all statements are processed.
+  reference accepts. A second `PARMS` statement also now adds to the first
+  rather than replacing it, matching the single field table `parmprc()` reads
+  once after all statements are processed.
 
   One job in the public `hazard` corpus changes, and only to drop a claim that
   was wrong: the second `%HAZARD` block of
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,11 +19,19 @@
   whether the keyword was present. Shape operands belonging to a phase that
   never activated are recorded in `$untranslated` rather than dropped, since
   `PROC HAZARD` zeroes them (`src/hazard/stmtprc.c`) and skips their covariates
-  (`src/hazard/setstat.c`). A `PARMS` that activates no phase at all is now
-  recorded too, because `PROC HAZARD` refuses that job outright
-  (`src/hazard/modterm.c`, `ERROR 1001: No phase selected`) where the
+  (`src/hazard/setstat.c`). A job that activates no phase at all is now
+  recorded too -- whether its `PARMS` named no positive `MU`, or it carried no
+  `PARMS` statement whatsoever -- because `PROC HAZARD` refuses that job
+  outright (`src/hazard/modterm.c`, `ERROR 1001: No phase selected`) where the
   translator would previously have emitted a runnable single-distribution fit
-  and reported full coverage. No job in the public `hazard` corpus is affected:
+  and reported full coverage. The two are one state rather than two cases:
+  `stmtprc.c` zeroes all three phases at initialization and only `setparmno()`
+  turns one back on, so a job with no `PARMS` has no active phase for the same
+  reason a `PARMS` naming `MUE=0` does. `modterm()` is reached on every job,
+  not only once a multiphase model has been selected -- its one call site in
+  `outmods()` is unconditional in the procedure's main sequence, and
+  `hazard.c` exits on `ERROR 1001` before any results are computed. No job in
+  the public `hazard` corpus is affected:
   the only `PARMS` statement in it that this gate changes is a documentation
   template with literal `?` placeholders, so this closes a latent divergence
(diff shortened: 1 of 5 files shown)


ehrlinger:
## Review findings — all addressed in `ff5c1c0`

The `r-reviewer` pass found four things. All four were verified against the reference C before acting, and all four are now fixed. Two were real defects I introduced; one was a wrong measurement in my own comment; one was the design tier, which is now upgraded.

### 1. (High) The widened guard fired on jobs `PROC HAZARD` runs — **fixed**

Dropping the `(length(mu) || length(early) || length(late))` qualifier changed the guard's meaning from *"PARMS specified something and no phase came of it"* to *"this parser did not build a phase"* — which silently also covers *"this parser could not **read** the operands."*

Verified, not taken on trust:

```sas
PROC HAZARD DATA=D; TIME FU; EVENT DEAD; PARMS MUE = 0.2 THALF = 1 NU = 1; RUN;
```

`.hzr_parse_hazard()` splits operands on `" "`, so this arrives as `"MUE" "=" "0.2" …` and nothing parses. But HAZARD's own lexer discards whitespace unconditionally — `hazard_l.l:32` `ws [ \t\n\r\)]+` with rule `{ws} ;` at `:50` — so `MUE = 0.2` is a well-formed `parmsopt` (`hazard_y.y:138`) and **that job runs with an active early phase**.

The refusal is a claim about the reference, so it is now gated on an `unreadable` flag set at each operand-level parse failure. `MUE=0 THALF=1` is fully readable, genuinely selects no phase, and still refuses — the gate keys on comprehension, not on `has_phases`.

This mattered more after the upgrade below: a wrong annotation became a wrong hard stop.

### 2. (High) Second false positive — multiple `PARMS` statements — **fixed**

```sas
PARMS MUE=0.2 THALF=1 NU=1;  PARMS FIXNU;
```

`parms_ops <- ops` **overwrote** rather than accumulated, so only `FIXNU` survived and the guard refused a job the reference runs. `PROC HAZARD` keeps one field table that `parmprc()` reads once after every statement is processed, so operands now accumulate.

Both false positives now verified clean, with a control:

| Input | Claims refusal |
|---|---|
| `PARMS MUE = 0.2 THALF = 1 NU = 1` (spaces) | `FALSE` ✅ |
| `PARMS MUE=0.2 …; PARMS FIXNU;` | `FALSE` ✅ |
| `PARMS MUE=0.2 THALF=1 NU=1` (control) | `FALSE` ✅ |
| `MUE=0 THALF=1` (true no-phase) | `TRUE` ✅ |

### 3. (Medium) My measurement was wrong — **dropped, not restated**

The comment claimed *"0 of the 29 `hazard()` fits lack `dist = "multiphase"`"*. That figure was measured on `main`, not on this branch — on #237's base it is **1** (`hm.dthar.TGA.sas`), which is the very job the same paragraph names as the one this gate touches. The sentence used as proof of latency the counterexample to itself. Removed rather than reworded.

### 4. (Low) Citation off by one — **fixed everywhere**

`hazard.c:298-301` → **`hazard.c:299-302`**. Line 298 is a bare `/* */`; the `1001` test is at 301 and `hzfxit("SEMANTIC")` at 302. Corrected in the comments, the tests, and the `$untranslated$reason` string — which callers grep, so it is data, not prose.

Every other citation in the diff was independently confirmed: `modterm.c:18-22`, `outmods.c:89`/`:91`, `hazard.c:296`, `stmtprc.c:87`, `setparmno.c:11`/`:14`, `parmprc.c:13,18,19` and `14-17,20-23`.

### 5. (Design) Record → **refuse**

Per maintainer decision. `.hzr_parse_parms()` now returns `refused`, and `.hzr_parse_job()` emits a `stop()` in place of the `hazard()` call, mirroring the `LCENSOR` + `ICENSOR` path at `sas-parse-job.R:536`. An `$untranslated` row alone left the fit chunk in place, so a reader who rendered past the callout still got a converged fit with a populated summary standing in for a job that produced nothing.

---

## ⚠️ One thing this changes in #237

Re-measured A/B across all 110 corpus `.sas` files: **exactly 1 of 57** translated jobs differs from #237 — and only by **dropping a false claim**.

`dist/examples/hm.dthar.TGA.sas` is a documentation template carrying literal `PARMS MUE=? THALF=? NU=? M=1 FIXM MUC=?;`. #237 currently annotates it *"PROC HAZARD selects no phase here and refuses the job (modterm.c:18-22 raises ERROR 1001)"*. That attributes the wrong refusal: the reference would reject `?` as a **syntax error**, long before `modterm.c` is reached. Its per-operand rows are unchanged, and **no corpus job is refused**.

## A test that could not have failed

The first version of the "no fit was emitted" assertion used `identical(x[[1]], quote(hazard))`. The translator emits `fit <- hazard(...)`, so the call head is `` `<-` `` and that assertion would have passed **unconditionally**. Caught by an unrelated failure in the same file. The helper now unwraps the assignment and the trap is documented where it bit.

## Gates

- `devtools::document()` — clean, no `man/`/`NAMESPACE` diff
- `lintr::lint_package()` — **0 lints**
- `devtools::test()`, `NOT_CRAN=true` — **0 failures**, 6 skips, **3789 passes** (skips unchanged from the AGENTS.md baseline, so nothing hid behind a new skip)
- `R CMD check --as-cran` from a clean `git archive` export, with the manual — **`Status: OK`**

On the check: the run before this commit reported `1 WARNING` on the PDF manual. It did **not** reproduce, on byte-identical `man/` inputs, and the `f19a3ec` base checked `OK` as well — so it was transient, not this branch. Recording it rather than omitting it.

ehrlinger:
## This PR also resolves the Codex P1 open against #237

A parallel session relayed a Codex P1 on #237 (head `f19a3ec`): `WEIBULL`, `FIXDELTA` and `DELTA=` are consumed without ever entering `mu`/`early`/`late`, so an operand-only `PARMS` slipped past #237's no-phase guard and emitted a runnable default-distribution fit at full coverage.

Verified against this branch's HEAD directly, rather than reasoning from the diff:

| Input | `has_phases` | `refused` | `$untranslated` |
|---|---|---|---|
| `PARMS WEIBULL` | `FALSE` | **`TRUE`** | 1 |
| `PARMS FIXDELTA` | `FALSE` | **`TRUE`** | 1 |
| `PARMS DELTA=0` | `FALSE` | **`TRUE`** | 1 |
| no `PARMS` | `FALSE` | **`TRUE`** | 1 |

End to end on a job whose only `PARMS` is `parms weibull;`, the emitted chunk is `fit <- stop("This PROC HAZARD job selects no phase: ...")` — **no `hazard()` call at all**.

So this branch covers the P1 and the wider no-PARMS case in one guard. The reason it does is that the gate here keys on **whether any phase is active**, with no qualifier about what `PARMS` happened to contain — which is what `modterm.c:18` actually tests:

```c
if(C->phase[1]==0 && C->phase[2]==0 && C->phase[3]==0)
```

That condition never asks whether a `PARMS` statement existed. The proposed alternative fix carved out `length(operands) > 0L` to preserve the no-PARMS path; that carve-out does not match the C, and is not in this branch.

### A note on test shape, which I agree with

`$coverage` counts whole statements and never `PARMS` operands, so it **cannot** register this defect and never could — `hzr_translate_sas()`'s own roxygen says coverage "is not evidence that the emitted calls execute." An `expect_gt(tokens_seen, tokens_mapped)` here would fail for a reason unrelated to the bug.

Worth stating plainly: my **first** commit's end-to-end test did assert `tokens_seen == tokens_mapped`. That test was rewritten when the behaviour became a refusal, so the assertion is gone — but it went for the right reason only incidentally, and I would not have caught it as a design error on its own. The current tests assert the absence of a `hazard()` call and the `$untranslated` row.

### Two branch-state facts for whoever merges

- `main` has moved to `f93d00c` (#236 merged). `f19a3ec` is **not** an ancestor of it. A merge of `origin/main` into this branch is clean; I have not performed it, because whether this PR stays stacked or collapses onto `main` is a maintainer decision.
- #237's own session has been idle since 00:44Z with that P1 open. I have not touched `fix/sas-parms-phase-keyed-on-mu`.

PR #239, opened 2026-09-09
fix(sas): start unspecified PARMS shape operands where PROC HAZARD does

ehrlinger:
## What

`.hzr_parse_parms()` filled any shape parameter a SAS `PARMS` statement did not name from **`hzr_phase()`'s** defaults. Three of those disagree with the reference C (`src/hazard/stmtprc.c:30-37`):

| | `NU` | `M` | `ETA` |
|---|---|---|---|
| `hzr_phase()` | 1 | 0 | 1 |
| `PROC HAZARD` | **2** | **1** | **2** |

The emitted call always carries `theta`, and `.hzr_optim_multiphase()` only falls back to `.hzr_phase_start()` when `theta_start` is `NULL` — so these were the values that actually reached the optimizer. The multiphase likelihood is multimodal, so a different start is a different answer, not a cosmetic difference.

## How the call and theta are kept in agreement

Switching only the theta table would have made the printed `hzr_phase()` call describe a fit that did not happen — worse than defaulting neither. So the fill happens **once**, in a new `.hzr_parms_fill_shape()`, and the same list builds both. A partially specified block now names every shape argument explicitly rather than leaving some to `hzr_phase()`'s formals. `.hzr_parms_theta_block()` no longer carries a default table at all — it asserts completeness, so a short list fails loudly instead of emitting a mis-aligned block.

## TAU

`TAU` cannot be reproduced at parse time: `stmtprc.c` sets it to zero and `SETG3()` (`setg3.c:317`) then derives `2*Tmax/3` from the data. Emitting `2 * max(<timevar>) / 3` was considered and rejected — SAS's `Tmax` is over the analysis set after exclusions, so the expression would look exact while being a guess. Such a phase is emitted at `tau = 1` and recorded in `$untranslated`. The predicate is the C one (`tau <= 0`, which also catches an explicit `TAU=0` that `hzr_phase()` would have rejected outright).

## The second defect, found in review

`SETG3_ignore_tau()` at `setg3.c:378-379` is **two** statements:

```c
Late.tau = ONE;
hzr_parm_set_fixed(HZ_TAU);
```

The first version of this PR read only the first, concluded "nothing to report", and shipped a test asserting `nrow($untranslated) == 0` on that branch. SAS fixes `TAU`; the emitted call left it free. At `alpha = 1` that is not a spare degree of freedom but an **unidentified** one — `G3` collapses to `(t/tau)^(gamma*eta)` (verified against `hzr_decompos_g3()` to 2e-16, now a test rather than a claim), so `mu` enters the likelihood only through `log_mu - gamma*eta*log_tau`. The translated fit converged onto that ridge and `vcov()` returned `NA` for both.

The emitted call now carries `tau = 1` and `"tau"` in `fixed=`, mirroring the C. On the test fixture that moves `log_mu`'s standard error from `NA` to 0.12 — asserted by fitting, not by inspecting the call.

⚠️ **This changes the emitted call for every job that fixes `ALPHA` at 1**, which is the dominant late-phase shape in the corpus. Deliberate, and flagged in `NEWS.md`.

## Two claims corrected

- "nothing previously translated changes" was **wrong**. `test-sas-translate-fits.R`'s end-to-end fit uses `MUE THALF NU MUC` with no `M` — a partial block. Its early phase moved from `m = 0` to `m = 1`. It needed no edit only because its assertions are on class, `dist` and theta *length*, none of which can see a start-value change.
- `MUL=0` was dropped from the "no positive MUL" test: that input emits `log(0) = -Inf` as its starting `log_mu` with no untranslated row, and asserting only that the TAU row is absent passed over the `-Inf` and quietly certified it. Tracked separately, not blessed here.

## Not in scope (reported, not folded in)

- `MUL=0` emitting `-Inf`, and shape operands with no `MUL` fabricating a whole phase `PROC HAZARD` would not build — both return zero untranslated rows. These belong with the phase-activation asymmetry the file already tracks.
- `mu`'s own default (`0.1` vs `SETG1`'s floor of 1) — same reason.

## Verification

- `devtools::document()` clean; `lintr::lint_package()` **0 lints**; `spelling::spell_check_package()` clean
- `devtools::test()` with `NOT_CRAN=true`: **0 failures**, 6 skips (SAS weighted-parity fixtures absent locally)
- `R CMD check --as-cran` from a clean `git archive`, with the manual: **Status: OK** (0/0/0)
- Two deliberate mutations confirmed the new call/theta agreement test can fail (8 and 7 failures)

One process note: the first `R CMD check` on the final commit reported `1 ERROR` with `getcwd: cannot access parent directories`. That was a collision on the shared per-user `$TMPDIR` — `AGENTS.md`'s recipe hardcodes `$TMPDIR/tree` and `$TMPDIR/TemporalHazard.Rcheck`, so a concurrent session running the same recipe deletes them underneath. Re-running in a unique directory gave `Status: OK`. That run's wall-clock is contended and is **not** quoted as a timing measurement; the uncontended runs earlier on this branch were 198s and 201s overall.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,19 +42,42 @@
   package's own fixture that moves `gamma`'s standard error from 7.5 to 0.02.
   This replaces an `$untranslated` row with a faithful translation.
 
-* **`hzr_translate_sas()` records, but deliberately does not mirror,
-  `SETG3_verify_ge_2()`.** When `GAMMA * ETA <= 2` and neither is fixed,
-  `setg3.c:907-921` rewrites the free one to `3` over the other -- and the
-  `PROC HAZARD` defaults `gamma = 1`, `eta = 2` land on that boundary exactly,
-  so it applies to every defaulted non-`WEIBULL` late phase. That constraint
-  exists to keep `PROC HAZARD` inside a numerical branch it can evaluate, the
-  same role `g3flag` plays; `hzr_decompos_g3()` carries the general
-  four-parameter form and needs no such restriction, and reaching a late shape
-  the reference implementation cannot is a purpose of this package. The
-  emitted call therefore keeps the general shape, and the divergence is
-  recorded so a SAS parity comparison knows why the starting values differ.
-  No corpus job is affected: every late-phase block there is `WEIBULL`, which
-  returns at `setg3.c:347` before this code runs.
+* **`hzr_translate_sas()` now reports the whole of what `SETG3()` would do to
+  a late phase, and mirrors only the part that has to be mirrored.**
+  `PROC HAZARD` does not optimize from the operands `PARMS` supplies: `SETG3()`
+  rewrites them first, and refuses some jobs outright. The translator now walks
+  that function (`src/model/setg3.c`) and records both, splitting them on a
+  single principle:
+
+  - A rewrite that resolves an **exact non-identifiability** is *mirrored*,
+    because the degeneracy is algebra and is just as real in R. There are two,
+    both in `SETG3_ignore_tau()`: the `TAU` pin and the `ETA` fix, described
+    above.
+  - Every other rewrite keeps `PROC HAZARD` inside a numerical branch it can
+    evaluate -- the role `g3flag` plays, which `hzr_decompos_g3()` does not
+    need because it carries the general four-parameter `G3` form. Copying those
+    would import a SAS limitation into R, and reaching a late shape the
+    reference implementation cannot is a purpose of this package. They are
+    *recorded* in `$untranslated`, so a SAS parity run knows why the starting
+    values differ.
+
+  In practice this covers `SETG3_verify_ge_2()`'s push of `GAMMA * ETA` clear
+  of 2 (which the `PROC HAZARD` defaults `gamma = 1`, `eta = 2` trip exactly,
+  so it applied to every defaulted non-`WEIBULL` late phase), the value
+  substitutions in all eight sign branches, and `SETG3_alpha_fixup()` /
+  `SETG3_alpha_gener()` deriving `ALPHA` from `GAMMA * ETA`.
+
+  Sixteen refusal codes are now recorded rather than emitted as runnable fits,
+  where previously only `SETG3980` was -- and that one only for `alpha = 0`,
+  not for the negative `ALPHA` that raises it too. `ALPHA = 0` **with**
+  `FIXALPHA` is the limiting exponential in both implementations and still
+  translates cleanly; left free, `SETG3` derives an `ALPHA` instead, which is
+  now reported.
+
+  No corpus job is affected: every late-phase block in the public corpus
+  carries `WEIBULL`, which returns at `setg3.c:347` before the sign dispatch,
+  and all of them write `TAU=1 FIXTAU ALPHA=1 FIXALPHA` with a positive `GAMMA`
+  and `ETA`. Verified by running the parser over all of them before and after.
 
 * **`hzr_translate_sas()` no longer discards the `ALPHA` and `ETA` a `PARMS`
   statement specified alongside `WEIBULL`.** The translator read the bare
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,9 +30,31 @@
   `$untranslated` to say so. The emitted call now carries `tau = 1` and
   `"tau"` in `fixed`, which identifies `log_mu` again. Jobs whose `PARMS`
   named a different `TAU` additionally record a row, since that value is used
-  by neither `PROC HAZARD` nor the translation. **This changes the emitted
-  call for every job that fixes `ALPHA` at 1**, which is the dominant late-
-  phase shape in the corpus.
+  by neither `PROC HAZARD` nor the translation. No corpus translation changes:
+  every late-phase block in the public corpus already writes `TAU=1 FIXTAU`,
+  so the emitted calls are byte-identical before and after (checked by running
+  the parser over all of them). What changes there is that those jobs no
+  longer need a warning.
+
+  The same branch's `ETA` fix (`setg3.c:405`) is mirrored too: with `GAMMA` and
+  `ETA` both free at `alpha = 1` the pair is exactly singular, and
+  `hzr_phase()` previously left both free and fitted the ridge. On the
+  package's own fixture that moves `gamma`'s standard error from 7.5 to 0.02.
+  This replaces an `$untranslated` row with a faithful translation.
+
+* **`hzr_translate_sas()` records, but deliberately does not mirror,
+  `SETG3_verify_ge_2()`.** When `GAMMA * ETA <= 2` and neither is fixed,
+  `setg3.c:907-921` rewrites the free one to `3` over the other -- and the
+  `PROC HAZARD` defaults `gamma = 1`, `eta = 2` land on that boundary exactly,
+  so it applies to every defaulted non-`WEIBULL` late phase. That constraint
+  exists to keep `PROC HAZARD` inside a numerical branch it can evaluate, the
+  same role `g3flag` plays; `hzr_decompos_g3()` carries the general
+  four-parameter form and needs no such restriction, and reaching a late shape
+  the reference implementation cannot is a purpose of this package. The
+  emitted call therefore keeps the general shape, and the divergence is
+  recorded so a SAS parity comparison knows why the starting values differ.
+  No corpus job is affected: every late-phase block there is `WEIBULL`, which
+  returns at `setg3.c:347` before this code runs.
 
 * **`hzr_translate_sas()` no longer discards the `ALPHA` and `ETA` a `PARMS`
   statement specified alongside `WEIBULL`.** The translator read the bare
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,23 @@
 
 ## Bug fixes
 
+* **`hzr_translate_sas()` now starts an unspecified shape parameter where
+  `PROC HAZARD` starts it, not where `hzr_phase()` does.** A `PARMS` statement
+  that named only some of a phase's shape operands had the rest filled from
+  `hzr_phase()`'s defaults, and three of them disagree with the reference C
+  (`src/hazard/stmtprc.c`): `NU` starts at 2 rather than 1, `M` at 1 rather
+  than 0, and `ETA` at 2 rather than 1. Since the multiphase likelihood is
+  multimodal, a different starting vector can reach a different optimum, so
+  this was a fidelity divergence rather than a cosmetic one. The defaults are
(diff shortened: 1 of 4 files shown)


ehrlinger:
## Update — P1 confirmed, and resolved by a principle rather than a blanket mirror

The finding is **correct**. `setg3.c:915` is literally `Late.gamma = 3.0e0/Late.eta;` — the `3` is a deliberate push clear of the boundary, not a normalisation *to* 2. And the new SAS defaults `gamma = 1, eta = 2` give `gte = 2` exactly, so this fires on **every** defaulted non-WEIBULL late phase. My branch emitted `gamma = 1` where PROC HAZARD starts at 1.5, silently — the defect class this PR exists to fix, reintroduced by the fix.

It is also broader than reported: `SETG3_ignore_tau()` does **not** return (`setg3.c:313-323`), so control falls through to the sign dispatch at `:359`. An `ignore_tau` job gets `verify_ge_2` too.

### Why `verify_ge_2` is recorded and not mirrored

The suggested remedy was to reproduce the normalization. I've deliberately not done that, and the reason is the point of the package rather than scope. `verify_ge_2`'s rewrite exists to keep PROC HAZARD inside a numerical branch it can evaluate — the same role `g3flag` plays, which this file already notes "has no R counterpart: `hzr_decompos_g3()` handles the general form directly". `hzr_phase("g3")` carries the general four-parameter shape and is under no such constraint, and **exercising a late shape the reference implementation cannot reach is a goal here**. Copying the rewrite would import a SAS limitation into R.

So it is recorded in `$untranslated`, because it remains a start-value divergence a SAS parity run needs explained.

### Where the C *is* mirrored, and why that is a different case

`setg3.c:405` fixes `ETA` when `GAMMA` and `ETA` are both free at `alpha = 1`. That is not a numerical-branch restriction — it is exact algebra: only the product enters the likelihood, so the pair is singular in R too. The previous commit removed the `(log_mu, log_tau)` ridge and left an identical one on `(gamma, eta)`, merely recorded. Now mirrored. Fitting both forms on the same fixture:

| | `gamma` SE |
|---|---|
| `ETA` free | 7.5 |
| `ETA` fixed, as `setg3.c:405` fixes it | **0.02** |

The test asserts the *ratio*, not just finiteness, so it cannot pass on a fixture change alone.

The value rewrite that follows (`gamma <- gamma*eta, eta <- 1`) stays unmirrored: likelihood-equivalent, and rewriting it would put the emitted call at odds with the user's own `PARMS` text.

### A correction to this PR's own description

I claimed the TAU pin "changes the emitted call for every job that fixes `ALPHA` at 1, which is the dominant late-phase shape in the corpus." **That is false.** Every late-phase block in the corpus already writes `TAU=1 FIXTAU`, so the emitted calls are byte-identical before and after — verified by running the parser over all of them under both revisions. It changes zero corpus emissions; only a warning disappeared. NEWS is corrected.

Relatedly, **no corpus job trips `verify_ge_2` either**: all late-phase blocks carry `WEIBULL`, which returns at `setg3.c:347` before the sign dispatch. Both issues are latent.

### Still open, not addressed here

The rest of the `SETG3` chain — `alpha_fixup`, the eight sign branches, and the refusal codes `SETG3900/910/920/930`, `SETG31010/31020`, `SETG31000/31040` — is neither mirrored nor recorded. Under the principle above most of those are "record", not "mirror", but they want their own change; `SETG3980` is currently the only refusal guarded.

Gates on the new head: `document()` clean, **0 lints**, spelling clean, `devtools::test()` **0 failures**, `R CMD check --as-cran` from a clean `git archive` with the manual **Status: OK** (245s, within the documented 199–245s spread).

ehrlinger:
## The rest of the `SETG3` chain, handled on the same split

`.hzr_setg3_notes()` now walks `src/model/setg3.c` in its own order and reports what `PROC HAZARD` would do to the late phase. It changes nothing about the emitted call.

**Mirrored** — only rewrites that resolve an *exact non-identifiability*, because that degeneracy is algebra and is just as real in R. There are exactly two, both in `SETG3_ignore_tau()`: the `TAU` pin and the `ETA` fix. I checked for a third by reading `hzr_decompos_g3()` at `alpha = 0` — `G3` is not a pure power of `t/tau` there and `tau` does not collapse into `mu`, so the limiting-exponential branch needs no mirror.

**Recorded** — everything else, because it keeps `PROC HAZARD` inside a numerical branch it can evaluate, the role `g3flag` plays. `hzr_decompos_g3()` carries the general four-parameter form and needs no such constraint, and reaching a late shape the reference cannot is a purpose of this package:

| `SETG3` element | |
|---|---|
| Entry refusals `SETG3900/910/920/930` (`:269-284`) | read the job's **own** `FIX*` flags, before `SETG3` sets any — so the trace snapshots `fixed_late` from before this branch's pins. `FIXTAU` with no `TAU` operand refuses on the `stmtprc.c` initializer of 0. |
| `verify_ge_2`'s push of `GAMMA*ETA` clear of 2 | the defaults `gamma = 1, eta = 2` trip it exactly |
| Value substitutions in all eight sign branches | |
| `alpha_fixup` / `alpha_gener` deriving `ALPHA` | |
| Sixteen refusal codes | against one before — and `SETG3980` now covers the **negative** `ALPHA` that raises it too, not just `alpha = 0` |

`ALPHA = 0` **with** `FIXALPHA` is the limiting exponential in both implementations and still translates cleanly; left free, `SETG3` derives an `ALPHA` instead, which is now reported. The pair is tested together so neither assertion can pass on a guard that fires unconditionally.

### Two subtleties the trace has to get right

- A **refused** job stops inside `SETG3` before either `TAU` rule runs, so those rows are suppressed rather than reported alongside the refusal — the same false-positive class the `MUL` gate exists to avoid.
- At `alpha = 1` only the product `gamma*eta` is identified, and the emitted call deliberately keeps the user's split rather than the C's. So the comparison is on the **product** there and on each operand otherwise; without that, every `ignore_tau` job would report a rewrite that changes no fit.

### Verification

- Corpus **unaffected**, checked by running the parser over every `PARMS` block before and after: byte-identical `$phases`, `$theta` and `$untranslated`. Every late-phase block there is `WEIBULL`, which returns at `setg3.c:347` before the sign dispatch.
- Four mutations of the new trace (drop an entry refusal; `3.0` → `2.0` in `verify_ge_2`; `alpha_fixup` never fires; WEIBULL stops skipping) produce **1, 1, 2 and 10** test failures.
- `document()` clean, **0 lints**, spelling clean, `devtools::test()` **0 failures**, `R CMD check --as-cran` from a clean `git archive` with the manual **Status: OK**.

One timing note, reported rather than restamped: two checks of the *same* tarball came in at **280s** and **251s** — a 29s spread on unchanged code, with the vignette step alone moving 58s → 46s. That straddles the top of the documented 199–245s band, so I can't separate real growth from machine variance on two points. Still far under CRAN's ~10 minute ceiling; worth a look next time the budget is measured properly.

### Still not covered

The `g_two` / `ga_two` halves of every predicate, driven by `FIXGE2` / `FIXGAE2`. Those tokens are unresolved by `.hzr_sas_token()` and already recorded as untranslated, so a job carrying either is never reported clean — the trace takes both as `FALSE` rather than guessing.

ehrlinger:
## Round 3: the rest of the SETG3 chain, a false refusal I introduced, and a merge with `main`

### The reviewer's headline finding was right, and it was mine to fix

`.hzr_setg3_notes()` refused jobs `PROC HAZARD` **runs**. `src/hazard/readobs.c:153-154` replaces an unspecified `TAU` with `0.75 * Tmax` on an active late phase, and `readobs()` runs at `hazard.c:276`, before `hzrg()` reaches `SETG3()` at `:292`. I verified the chain rather than taking it: `parmprc.c:20` gives `TAU` ispec index 8, and `setprmf.c:13-16` sets `ispec(8) = 2` when the operand is absent — exactly `readobs.c`'s condition.

Two errors followed from reading `stmtprc.c` without `readobs.c`:

- A bare `FIXTAU` was recorded as `SETG3900`, *"PROC HAZARD refuses this job"*. It does not — it fits with `TAU` held at `0.75 * Tmax`. **An `$untranslated` refusal for a job that runs is this package's signature defect inverted**, and two of my new tests asserted it.
- The `2*Tmax/3` row (`setg3.c:317`) named the wrong rule and constant for an absent `TAU`. That branch is reachable only through a `TAU` the job *wrote* as non-positive. Pre-existing, inherited here.

`tau_absent` and `tau_nonpositive` are now separate; only the second can refuse.

### Also from that review

- `.hzr_setg3_refusal_reason()`'s `is.null()` fallback was **hollow**: `[[` on an unmatched name in a character vector throws rather than returning `NULL`, so the parser would have errored instead of degrading.
- TAU-row suppression is keyed on whether the refusal is an **entry** refusal (`setg3.c:269-284` — the only four raised before the TAU rules), not on "refused at all". The other twelve fire from `:429`, after the TAU rule has run.
- The rewrite row claimed the call "keeps the general shape deliberately" even where the operand is outside `hzr_phase()`'s domain, so the call cannot be built at all. That case now says so, and the test asserts the error instead of describing it in a comment.

### Where I disagreed, with evidence

The review asked for tests on eight refusal codes. **Seven are unreachable in both languages** — each guards a condition the entry checks already refused. Established by exhaustive grid over `(tau, gamma, alpha, eta, fixed-subset, weibull)` rather than argued through three files, which is how the `SETG3900` claim went wrong in the first place. Nine codes can fire; NEWS said sixteen and now says nine.

### Merge with `main`

#237 (MU-keyed phase gate) and #238 (refuse a no-phase job) landed in the same function. Resolved: `main`'s gates and `refused` path, this branch's PROC HAZARD shape defaults and filled shape lists, one `mu_active()` definition rather than two, and `main`'s `ALPHA=1 GAMMA*ETA` guard dropped because this branch mirrors `setg3.c:407` instead of recording it.

One genuine interaction, not a test artifact: an unspecified late phase defaults to `gamma = 1, eta = 2`, product exactly 2, so `SETG3_verify_ge_2()` fires on fixtures `main`'s tests expected to be clean. Those tests now expect it.

### Verification

The parser's output over **every corpus `PARMS` block is byte-identical between `origin/main` and this branch** — `$phases`, `$theta` and `$untranslated`, 26 blocks. The whole branch still changes no corpus translation.

`document()` clean · **0 lints** · spelling clean · `devtools::test()` **0 failures** · `R CMD check --as-cran` from a clean `git archive` with the manual: **Status: OK**, 222s (inside the documented 199–245s spread).

### Still open, deliberately

`SETG3950` is not recorded (data-dependent, undecidable from a `PARMS` block) and the `g_two`/`ga_two` codes are out of scope because `FIXGE2`/`FIXGAE2` are already recorded as unresolved. Both are documented in the table rather than left silent.

PR #240, opened 2026-09-09
docs(agents): restamp the check budget at bd75f9f

ehrlinger:
`AGENTS.md` asks for the check-time figures to be re-measured and the commit stamped when they change. The 1.2.10 cycle (#237, #238) moved them; the table was still quoting `fe57e60` from 2026-09-05.

## Measured 2026-09-09 at `bd75f9f`

Tarball built from a clean `git archive` export, checked with the manual. `Status: OK`.

| | Now | Was (`fe57e60`) |
|---|---|---|
| Overall | **3m 41s** (221s) | 3m 32s (212s) |
| Tests | 90s / 96s | 87s / 93s |
| Vignette rebuild | 47s / 52s | 42s / 46s |
| Tarball | 2.93 MB | 2.91 MB |
| Under check | 193 skips / **2765** passes | 193 / 2679 |
| Local | 6 skips / **3796** passes | 6 / 3703 |
| `skip_on_cran()` calls | 173 | 173 (unchanged) |

Overall is from two independent full runs — the first attempt's `zsh` `time` plumbing failed to capture wall time, so it was re-run; the component timings agreed across both (tests 90s each). Still a wide margin against CRAN's ~10 minute ceiling.

## The figure that actually matters here is not the 9 seconds

Against the 46-second spread this file documents on *unchanged* code, a 9s delta is well within noise on its own. What makes it readable is that **the skip count held at 193 and `skip_on_cran()` held at 173 while under-check passes rose by 86.** This cycle's new tests run *inside* the check rather than being skipped past it.

That is the exact opposite of the 2026-09-05 cycle, where the row fell because five end-to-end `hazard()` fits were added carrying `skip_on_cran()` — the check measuring less work, not the same work faster.

So the ⚠️ paragraph that read "the 2026-09-05 row went DOWN and that is not a speedup" is now a pair covering both directions, with the lesson generalised: **seconds alone cannot distinguish a cheaper suite from a smaller one**, so read this column against the under-check pass count in "Definition of done".

The "Definition of done" gap figure is updated to 1,031 passes (from 1,024), with a note that it barely moved this cycle while the suite grew by 93 — for the same reason.

## Scope

`AGENTS.md` only. No code, no tests, no `man/`. `lintr` 0 lints, `spelling` 0 rows.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,13 +47,15 @@ Four details there are load bearing:
 - **Lint runs before tests** because it is seconds against about a minute for the suite. Cheap
   failures first.
 - **`R CMD check` does not run the whole suite.** It runs with `NOT_CRAN` false, so
-  `skip_on_cran()` tests are skipped: 193 skips and 2679 passes under check, against 6 skips
-  and **3703** passes locally (both measured 2026-09-05 at `fe57e60`, the `v1.2.9` release
-  artifact, with no environment variables set; the local figure needs the SAS fixture
-  checkouts present under `~/Documents/GitHub/hazard`). **That gap is now 1,024 passes**, up
-  from about 900 when it was last measured on 2026-08-31, so the check covers proportionally
-  less of the suite than it did. A green check is **not**
-  evidence that those tests pass; only the local `devtools::test()` line is.
+  `skip_on_cran()` tests are skipped: 193 skips and 2765 passes under check, against 6 skips
+  and **3796** passes locally (both measured 2026-09-09 at `bd75f9f`, with no environment
+  variables set; the local figure needs the SAS fixture checkouts present under
+  `~/Documents/GitHub/hazard`). **That gap is 1,031 passes**, against 1,024 on 2026-09-05 and
+  about 900 on 2026-08-31. Note the gap barely moved this cycle while the suite grew by 93
+  passes: the skip count held at 193 and `skip_on_cran()` calls held at 173, so this cycle's
+  new tests run *under* the check rather than being skipped out of it. That is the opposite of
+  the previous cycle and the reason the timing rows below move the way they do. A green check
+  is **not** evidence that the skipped tests pass; only the local `devtools::test()` line is.
 - **Commit before you `git archive`.** It exports the committed tree, so an uncommitted fix is
   silently absent and the check answers a question about the wrong code. This has already
   cost one wrong conclusion. Nothing in the output tells you.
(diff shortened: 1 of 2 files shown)


ehrlinger:
## Added: the parity inventory restamp (`c2e27ec`)

Re-derived using the file's own documented command rather than tallying. Two things were wrong — **neither of them the per-test counts**.

### 1. The re-derivation command never produced the documented number

The file instructed:

```r
as.data.frame(testthat::test_local(filter = "sas-parity"))[, c("test", "passed")]
```

`filter` is a **regex**, so that also matches `test-bhdead-sas-parity.R` (3 tests) and `test-weights-sas-parity.R` (4 tests):

| filter | files matched | tests | expectations |
|---|---|---|---|
| `"sas-parity"` | 3 | 25 | **196** |
| `"^sas-parity$"` | 1 | 18 | **194** |

196 reconciles with neither the documented **63** nor the file's own **194**. Anyone following the instruction got a number matching nothing in the document and no way to see why. Anchored, and verified to isolate `test-sas-parity.R`.

### 2. "63/63 PASS" reads as the file total and is not

It is the **six prototype fits** listed beneath it. `test-sas-parity.R` now holds **18 tests / 194 expectations**, 0 failing. 63 *was* the file total in May 2026 and hasn't been since — the section heading says "Prototype status (2026-05-12)", but the inline "measured 2026-09-08" note invited reading it as current.

Now scoped explicitly, with the file total stated alongside so the two numbers can't be confused again.

### What did NOT change

All six per-test counts re-derive **exactly** as documented — 14, 11, 12, 14, 9, 3 = 63. Nothing about those fits drifted. Only the framing around them was wrong, and the commit says so rather than implying a correction that didn't happen.

Measured 2026-09-09 at `bd75f9f`. `lintr` 0 lints, `spelling` 0 rows. No code changed.

This PR is now two commits: the `AGENTS.md` check budget and this. Both are dated-measurement restamps, so they belong together; say the word if you'd rather they were separate PRs.

ehrlinger:
## [P2] Separate coverage from timing — addressed in `9d4c6fa`

Both points are correct, and both were mine. Verified against the numbers before changing anything.

### 1. "the check measured less work than before" was unsupported

Right. Adding tests that carry `skip_on_cran()` does not reduce what the check runs — those tests never ran under it either way. The check's workload is unchanged; what fell is the **proportion** of the suite it covers. Restated as a coverage claim.

(I preserved that sentence from the previous text while rewriting around it, rather than authoring it — but I rewrote the paragraph and had every opportunity to check it, so it is mine now.)

### 2. The 9-second increase is not made "real" by the assertion count

Also right, and this one I introduced. `2679 → 2765` proves the check's workload grew; it does not license attributing a 9s delta to it when the same section documents a **46-second spread on effectively unchanged code**. I then hedged in the very next sentence instead of just not making the claim — which is worse than making it cleanly, because the bolded line is what a reader carries away.

Timing is now recorded as an observation and explicitly not explained.

### One thing that sharpens the finding rather than just conceding it

The 2026-09-05 "fall" that the old paragraph built its causal story on was **213s → 212s**. One second. Against a 46-second spread, there was nothing there to explain. That is now stated, because it is the most concrete illustration in the file of why these two series must be read apart.

### Shape of the fix

- **What the assertion counts establish (coverage):** 09-05 coverage fell proportionally; 09-09 coverage largely held.
- **What the seconds do NOT establish (timing):** 212s → 221s and 87s → 90s, both far inside the spread, attributable to nothing.
- Closing note on why both series are kept: seconds say whether CRAN's ceiling is in danger; the under-check pass count says how much of the suite the check runs. Only the second can distinguish a cheaper suite from a smaller one, and it cannot be read off the clock.

The cross-reference in "Definition of done" no longer claims the coverage figures explain the timing rows.

`lintr` 0 lints, `spelling` 0 rows. No code changed. Keeping both restamps in this one PR per the maintainer.

ehrlinger:
## Merge plan: this PR now, numbers restamped after #239

Recorded so the decision is visible to whoever reads this later, and so the stale figures are a known follow-up rather than a surprise.

**Merging this before #239 knowingly dates part of it.** #239 (`fix/sas-parms-sas-shape-defaults`) adds roughly 400 lines of tests to `test-sas-parse-parms.R` and `test-sas-parse-hazard.R`. That will move exactly the figures stamped here:

| Stamped here at `bd75f9f` | Moves when #239 lands |
|---|---|
| local 3796 passes | yes |
| under-check 2765 passes | yes, unless the new tests are skipped |
| tests 90s / overall 221s | probably, slightly |

**What does NOT go stale**, and is the reason to merge now rather than wait:

- the coverage-vs-timing separation (the [P2] fix in `9d4c6fa`) — a methodology correction, independent of any measurement;
- `filter = "^sas-parity$"` in the parity inventory — the unanchored form never reproduced any number in that document;
- the scoping of "63/63" to the six prototype fits rather than the whole file.

**Follow-up owed after #239 merges:** re-run `R CMD check --as-cran` from a clean `git archive` export plus the local suite at the new `main` head, and update the four figures above — the check-budget table row, the component table, the "Definition of done" pass counts, and the parity inventory's file total. One commit, no methodology changes.

Note #239 is currently `DIRTY`: its merge-base predates #237 and #238, and all three rewrite the same region of `R/sas-parse-parms.R`. It replaces the parser's shape defaults with `PROC HAZARD`'s own (`nu` 1→2, `m` 0→1, `eta` 1→2) and rewrites the guard-(4) comment #238 added. Worth confirming the `refused` and `unreadable` gates still behave once those defaults change — the tests covering them live in the same file #239 rewrites.

PR #249, opened 2026-09-11
ci: run the SAS parity and translator corpus tests on ubuntu-release

ehrlinger:
## What

The **ubuntu-latest / release** job in `R-CMD-check.yaml` now checks out the public SAS reference corpus, [`ehrlinger/hazard`](https://github.com/ehrlinger/hazard), pinned at `dad7978`. It points `HAZARD_REPO` and `HAZARD_EXAMPLES_DIR` at it, so the SAS parity tests and the translator corpus test run on CI instead of skipping.

Until now they ran only on a machine with `~/Documents/GitHub/hazard`. A green CI run said nothing about them, and the corpus test is the one that caught #243's translator issue.

## Design

- **One runner only.** The added tests take about 110s (`test-sas-parity.R`) plus 23–30s (the corpus test) locally, and more on a runner. Windows is already the slowest job, at about 45 min. ubuntu-release is already a required check, so these tests now block merges with no ruleset change.
- **Pinned to a commit.** A change in `hazard` can't move this package's CI unannounced. Moving the pin is a deliberate PR, because the corpus test's floors are measured against it.
- **Checked out beside the package, then moved to `$RUNNER_TEMP`.** Anything left in the package root would be packed into the tarball by `R CMD build`.
- **Fails loudly if the fixtures are missing.** A step checks for `examples/hz.death.AVC.lst`, `hz.deadp.KUL.lst` and `hz.te123.OMC.sas` and fails the job if any is absent. A moved path therefore turns CI red instead of back into silent, green skips.
- No `pull_request_target` or event data is involved. `ref` is a literal SHA, and `persist-credentials: false` is set.

## What the local proof found: the corpus test was measuring a local build artefact

Before pushing, I ran the four affected test files against a **fresh clone at the pin**. `HOME` pointed at an empty directory, so the `~/Documents/GitHub/hazard` default couldn't satisfy anything. The env vars work: the tests read the clone, and the control run without them skips.

**But the corpus test failed** (`n_partial` 11 < 12). The local checkout matches `dad7978` according to `git status`, but it also holds **`dist/`, a gitignored `make install --prefix=./dist` tree from April** with 44 extra `.sas` files. The corpus test lists `.sas` files recursively, so its floors had been measured over files that exist only on this machine:

| Tree | `.sas` files | translations | distinct | eligible | partial | rendered |
|---|---|---|---|---|---|---|
| clean clone at `dad7978` (what CI sees) | 66 | 37 | 21 | 10 | 11 | 10 |
| local, before this PR (`dist/` counted) | 110 | 57 | 22 | 10 | **12** | 10 |
| local, with this PR (`dist/` excluded) | 68 | 39 | 21 | 10 | 11 | 10 |

**Correction to #243.** The corpus job that #243 turned from "rendered" into "partial", `hm.dthar.TGA.sas`, **exists only in that `dist/` tree**. It isn't tracked anywhere in `hazard`. The comment I wrote in #243 cited it as corpus evidence, which was wrong. The translator fix is still tested on CI, without the corpus, in `test-sas-translate-fits.R`.

**Fix in `test-sas-translate-corpus.R`:**
- Both tests exclude the top-level `dist/`, so a local run and CI measure the same corpus (the distinct/eligible/partial/rendered counts now match).
- The partial floor is 11, the clean-tree count.
- The measurement comments are corrected and dated.

The test now passes on all three trees: 142 passes on the clean pin, 146 locally, 0 failures.

## Coverage: before and after

On the latest green `main` run (`34517773078`), ubuntu-release skipped 44–45 tests against 6 locally. This PR should recover **21** of them:
- 17 in `test-sas-parity.R`;
- 3 "hazard checkout not available" (the corpus and grammar tests);
- 1 `hz.te123.OMC.sas` round-trip.

What stays CI-only, because no runner can satisfy it:
- the `bh.dead` helpers (they live in `inst/dev/`, which the built tarball excludes);
- the C `hazard` binary;
- the private maze datasets;
- the generated weighted-parity fixture.

`AGENTS.md` describes the mechanism. I'll stamp the measured ubuntu-release skip count there from this PR's own CI run.

## Gates

| Gate | Result |
|---|---|
| `actionlint` on `R-CMD-check.yaml` | clean |
| `lintr::lint()` on the changed test | 0 lints |
| Corpus test vs a fresh clone at `dad7978`, with `HOME` redirected so the default path cannot help | 0 failed, 142 passed |
| Corpus test vs the local checkout (with its `dist/`) | 0 failed, 146 passed |
| All four `HAZARD_REPO` / `HAZARD_EXAMPLES_DIR` test files vs the fresh clone | read the clone; the control without the env vars skips |
| `NOT_CRAN=true devtools::test()` | **0 failed, 6 skipped, 4149 passed** |

**Not run, and why:**
- **`devtools::document()`:** no roxygen changed.
- **`spelling`:** it doesn't cover `AGENTS.md`.
- **Local `R CMD check --as-cran`:** the only tarball change is `tests/testthat/test-sas-translate-corpus.R`, and that file is skipped under `--as-cran` (`skip_on_cran()`). `.github/` and `AGENTS.md` are `.Rbuildignore`d.

The real evidence for this PR is its own CI run: ubuntu-release has to show the SAS tests running, not skipping.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -61,6 +61,11 @@ jobs:
           fi
 
   R-CMD-check:
+    # Named explicitly: without this, GitHub appends every matrix value to the
+    # job name, so sas_corpus turns the ubuntu-release job into
+    # "R-CMD-check (ubuntu-latest, release, true)" and the ruleset's required
+    # "R-CMD-check (ubuntu-latest, release)" never reports.
+    name: R-CMD-check (${{ matrix.config.os }}, ${{ matrix.config.r }})
     needs: changes
     runs-on: ${{ matrix.config.os }}
     strategy:
@@ -101,13 +106,21 @@ jobs:
           persist-credentials: false
 
       # Fail here if the files the tests look for are not where they look,
-      # rather than letting every one of those tests skip and pass.
+      # rather than letting every one of those tests skip and pass. The list
+      # is every fixture test-sas-parity.R reads, because each test there
+      # skips on its own file: a pin that keeps a few sentinels but drops one
+      # of these would turn that test green over nothing.
       - if: needs.changes.outputs.run == 'true' && matrix.config.sas_corpus
         shell: bash
         run: |
           mv hazard-corpus "$RUNNER_TEMP/hazard"
           for f in examples/hz.death.AVC.lst examples/hz.deadp.KUL.lst \
-                   examples/hz.te123.OMC.sas; do
+                   examples/hm.death.patient.lst examples/hm.death.AVC.deciles.lst \
+                   examples/ac.death.AVC.lst examples/hp.death.AVC.lst \
+                   examples/bs.death.AVC.lst examples/hp.death.AVC.hm1.lst \
+                   examples/hp.death.AVC.hm2.lst examples/hs.death.AVC.hm1.lst \
+                   examples/hz.te123.OMC.lst examples/hz.tm123.OMC.lst \
+                   examples/hz.te123.OMC.sas examples/data/omc; do
             test -f "$RUNNER_TEMP/hazard/$f" || { echo "missing: $f"; exit 1; }
           done
           echo "HAZARD_REPO=$RUNNER_TEMP/hazard" >> "$GITHUB_ENV"
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -68,7 +68,7 @@ jobs:
       matrix:
         config:
           - {os: ubuntu-latest, r: 'devel'}
-          - {os: ubuntu-latest, r: 'release'}
+          - {os: ubuntu-latest, r: 'release', sas_corpus: true}
           - {os: ubuntu-latest, r: 'oldrel-1'}
           - {os: macos-latest,  r: 'release'}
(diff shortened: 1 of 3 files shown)


ehrlinger:
Copilot review follow-up, covering the comment suppressed in the review body (the sentinel list is incomplete).

At the pin (`dad7978`) all eight `.lst` fixtures that the enabled parity tests `skip_if_not()` on are present, so the run is not hollow today. A later pin move could make it hollow without any warning, though. Accepted: the sentinel step will check every fixture the enabled tests gate on. This will be pushed after #250 merges.

ehrlinger:
Another finding, and the reason this PR shows BLOCKED with every check green: adding `sas_corpus: true` to the matrix renamed the ubuntu-release job to `R-CMD-check (ubuntu-latest, release, true)`, because GitHub appends every matrix value to an unnamed job. The `protect main` ruleset requires `R-CMD-check (ubuntu-latest, release)`, which now never reports. Merging this as it stands would leave every later PR BLOCKED too.

Fixed locally in `55cfd98` by naming the job explicitly, `R-CMD-check (${{ matrix.config.os }}, ${{ matrix.config.r }})`. The same commit extends the sentinel step to all fourteen fixtures. It will be pushed after #250 merges; once CI runs, confirm that the required context reports under the old name.

PR #306, opened 2026-09-14
fix(stepwise): test an entering candidate on the column its refit added (#305)

### LINKED ISSUE #304: Backward hzr_stepwise() errors when theta is named by covariate ('Unknown coefficient name(s): beta1')
Backward `hzr_stepwise()` stops with an error when the base fit's `theta` names its covariate coefficients after the covariates (for example `age`, `mal`). The drop path looks coefficients up by a positional `beta1`, `beta2`, ... name. This fails loudly; it does not give a wrong answer.

## Reproducer (confirmed on `main` at `7e50e2b`, 2026-09-13)

```r
devtools::load_all()
d <- na.omit(avc[, c("int_dead", "dead", "age", "mal")])
run <- function(th) {
  f <- hazard(survival::Surv(int_dead, dead) ~ age + mal, data = d, dist = "weibull", theta = th, fit = TRUE)
  hzr_stepwise(f, data = d, direction = "backward", criterion = "wald", slstay = 0.2, trace = FALSE)
}
run(c(mu = 0.1, nu = 1, age = 0, mal = 0))
# Error: Unknown coefficient name(s): 'beta1'.  Available: 'mu', 'nu', 'age', 'mal'
```

| `theta` naming | `names(coef(fit))` | backward `hzr_stepwise()` |
|---|---|---|
| unnamed covariate starts, `c(mu, nu, 0, 0)` | `mu, nu, "", ""` | runs |
| **named by covariate**, `c(mu, nu, age = 0, mal = 0)` | `mu, nu, age, mal` | **error** |
| named `beta1`/`beta2` | `mu, nu, beta1, beta2` | runs |

Naming `theta` after the covariates is the natural choice. It also avoids the empty coefficient names that unnamed starting values produce. So the case that errors is the one a user is most likely to write.

## Fix

The drop path should identify a covariate's coefficient by the design column it belongs to, not by a positional `beta<k>` name. Whatever `theta` is named, the refit should then work. A test should cover all three namings above, and should also check that the dropped variable and the step's p-value are the same under each naming.

Found in passing by the session working on #302 (stepwise refit-failure reasons).

### LINKED ISSUE #305: Forward stepwise Wald test can test the wrong coefficient when a candidate's name collides with a factor dummy
The forward stepwise **Wald** test can test the **wrong coefficient** for an entering candidate. It then accepts or rejects the step on that coefficient's p-value. This is a silent wrong answer.

## Cause

`.hzr_candidate_coef_name()` looks up an entering candidate's coefficient **by the variable's bare name**. `model.matrix()` names a logical `flag`'s column `flagTRUE`. If a factor `fla` (levels `f`, `g`) is already in the model, its dummy column is named `flag`. So the lookup returns the factor's dummy, and that test is reported as the candidate's. Two call sites are affected: the Wald candidate loop in `.hzr_stepwise_forward_step()`, and the score path's Wald fallback (#130). Both the single-distribution and multiphase paths are affected.

## Reproducer (confirmed on `main` at `7e50e2b`, 2026-09-13)

```r
devtools::load_all()
n <- 400; set.seed(11)
d <- data.frame(fla = factor(sample(c("f", "g"), n, TRUE)), flag = sample(c(TRUE, FALSE), n, TRUE))
d$time <- rexp(n) * exp(-1.2 * d$flag); d$status <- 1
base <- hazard(survival::Surv(time, status) ~ fla, data = d, dist = "weibull", theta = c(.5, 1, 0), fit = TRUE)
.hzr_stepwise_forward_step(base, scope = ~ flag, criterion = "wald", slentry = 0.05, data = d)$accepted
# FALSE
```

**Truth:** fit `~ fla + flag` directly. `flagTRUE` has z = 11.39, p = 4.68e-30, so the step should be accepted. The `fla` dummy (named `flag`) has p = 0.163, which is what the step reports. The session that found this also measured the multiphase score path's Wald fallback: it reported z = -1.35, p = 0.178, "not entered", for a candidate whose real z was 50.3.

## Fix

Identify the entering coefficient as the design column that the refit has and the current fit lacks, and require exactly one such column, as the score path already does. A fix is in progress on `fix/stepwise-candidate-coef-name` (draft PR to follow). The same name collision on the **backward** step is recorded on #304, which has the same root cause.

ehrlinger:
## Summary

`.hzr_candidate_coef_name()` looked up an entering candidate's coefficient by the variable's bare name. That picks the wrong coefficient whenever another term's column already has that name. `model.matrix()` names a logical `flag`'s column `flagTRUE`, while a factor `fla` with level `g` owns a column named `flag`. With both in play, the lookup returned **the factor's dummy**. The Wald candidate loop and the score path's Wald fallback then reported that dummy's test as the candidate's p-value, with no error and no warning. Single-distribution and multiphase fits were both affected.

**Reproducer** (n = 400, `set.seed(11)`, logical `flag` with true effect 1.2, base `~ fla`):

| Path | Reported for `flag` | The candidate's own test |
|---|---|---|
| single-dist, Wald | `beta1`, p = 0.163 | `beta2` (`flagTRUE`), p = 4.7e-30 |
| multiphase, Wald | `constant.flag`, p = 0.178 | `constant.flagTRUE`, z = 12.4 |
| multiphase, score → Wald fallback (effect 5) | z = −1.35, p = 0.178, **not entered** | z = 50.3 |

## Fix

- Both forward-step call sites now pass the current fit (`current = current`). The new `.hzr_entered_coef_name()` identifies the candidate as the design column that the refit has and the current fit lacks. That is the rule `.hzr_score_expand_multiphase()` already uses to place its pinned zero (`R/score-test.R`, "The candidate's column is not always the phase's last").
- The resolver also requires the design to gain **exactly one column**, as the score path does. Without that check, adding `z` to a phase `~ z:f` turns `z:fa, z:fb` into `z, z:fb`. That is one new name, the same column space and an identical objective (−291.8985 for both), yet the step reported p = 8.9e-10 and entered `z`. This is pre-existing: the old name lookup accepted it too. It is now an error.
- A candidate that adds more than one column still errors "expands to multiple coefficients". Duplicate column names (no new name) error instead of guessing.

## Behaviour change

Single-distribution fits now accept one-column logical and two-level factor candidates under `criterion = "wald"`. Before, those stopped with "not found in the design matrix". Yet the score criterion's refusal message names `criterion = "wald"` as the way to test such a column, and multiphase fits already accepted them. `test-stepwise-driver.R` required the old error; it now asserts that the factor is tested on its own column (`faclo`), and the score refusal is unchanged.

## Tests

New file `tests/testthat/test-stepwise-candidate-coef-name.R`, seven tests:
- Five end-to-end tests compare the reported `stat` with a z computed from a direct fit's `coef()` and `vcov()`. Each first asserts that the candidate's z and the competing column's z are more than 5 apart, so a match is evidence.
- The fallback test asserts that the fixture reaches the fallback (`fallback == TRUE`, `stat_type == "wald_z"`).
- One test puts the candidate third of four columns (`~ a * b` plus `c`).
- The column-count test pins its mechanism (column names, equal objective) before it asserts the error.

**Mutation checks.** Each ran in its own scratch copy, after `grep -c` confirmed the mutation had landed:
- `current = current` removed at the Wald-loop site: tests 1–3 fail; test 2 errors with the old "not found".
- `current = current` removed at the fallback site: the fallback test fails, and nothing else does.
- The resolver changed to return the last column: the non-last-column test fails, and nothing else does.
- The column-count guard absent (the RED run): the reparameterisation test fails.

### On the head, merge commit `6fe307a` (merged `origin/main` at `8bf0681`, which includes #289)

This merge brings in exactly #289's three files: `NEWS.md`, `R/likelihood-multiphase.R` and `tests/testthat/test-predict-hazard-multiphase.R`. None of them is a stepwise file. The only conflict was one `NEWS.md` hunk, and I kept every bullet: each appears exactly once, #289's "unnamed vector" bullet included. My changes to `R/stepwise-step.R` are intact (both call sites, the helper, the signature and the column-count guard).

- `devtools::document()`: no changes
- `lintr::lint_package()`: 0
- `spelling::spell_check_package()`: 0
- `NOT_CRAN=true devtools::test()`: **5582 passed, 0 failed, 0 errors**, 655 warnings, 1584 tests. The **6 skips** are all the kind that always skip locally: 3 bh.dead SAS fixture, 1 C binary (status 126), 2 weighted-parity fixture. No test skipped for lack of the study volume. Other sessions' R jobs were running at the same time, none of them in this worktree, so no timing is given. This is the first run that combines #289's `unname()` change with this fix.
- `R CMD check --as-cran`: not re-run on this commit, at the queue coordinator's direction, because this PR's own code is unchanged since the check on `8c5a972` below. CI runs the five-platform R-CMD-check on this head.

### On the merge commit `8c5a972` (merged `origin/main` at `9ec83e7`, which includes #292)

The only conflict was one hunk in `NEWS.md`, where both sides added bullets at the top of "Bug fixes". I kept all of them, and each appears exactly once. My changes to `R/stepwise-step.R` came through intact: both `current = current` call sites, `.hzr_entered_coef_name()`, the `current = NULL` signature and the column-count guard. `main` changed none of the stepwise files since `7e50e2b`.

- `devtools::document()`: no changes
- `lintr::lint_package()`: 0
- `spelling::spell_check_package()`: 0
- `NOT_CRAN=true devtools::test()`, run on its own with `/Volumes/qhsstudies` mounted: **5550 passed, 0 failed, 0 errors**, 650 warnings, 1583 tests. **6 skips, all of the kind that always skip locally:** 3 for the missing bh.dead SAS fixture, 1 because the C binary exits with status 126, and 2 for the missing weighted-parity fixture. There were no volume skips, so the maze and achalasia `%repeat` parity tests ran.
- `R CMD check --as-cran` with the manual, from `git archive` into a fresh `mktemp -d`: **Status: OK**. Tarball 3,276,933 bytes, with no developer files in it. Other sessions' R jobs were running during the check, so its wall time is not a timing to record.

### Before the merge, at `aef3f4c`

- `devtools::document()`: no changes to `man/` or `NAMESPACE`
- `lintr::lint_package()`: 0 lints
- `spelling::spell_check_package()`: 0
- `NOT_CRAN=true devtools::test()` at `aef3f4c`: **5181 passed, 0 failed, 0 errors**, 647 warnings, 1517 tests. There were 12 skips, each with a known cause:
  - **6 volume skips.** `/Volumes/qhsstudies` is down: 5 are "maze study datasets not mounted" in `test-repeated-events-parity.R`, and 1 is "achalasia study datasets not mounted" in `test-repeated-events-renewal-parity.R`.
  - **The 6 skips that are always there locally.** 3 are "bh.dead SAS fixture not available", 1 is the C binary ("Legacy binary returned non-zero status 126"), and 2 are "Weighted parity fixture weighted-avc-fractional.rds not found".
- `R CMD check --as-cran` (with manual, from `git archive` of `aef3f4c`): **Status: OK**. The tarball is 3,251,241 bytes (`stat -f %z`), with no `CLAUDE`, `AGENTS` or `reports/` files in it. The check took 4m43s, but the full suite ran on the same machine at the same time, so that is not a timing to record. The branch head, `407bab0`, adds only the `(#305)` citation to NEWS on top of `aef3f4c`, and no check has run on it yet. The gates will run again after `origin/main` is merged in, before this is marked ready.
- `r-reviewer`: found no silent wrong answer introduced by the diff. Both of its findings, the column-count case and the missing non-last-column test, are fixed in this PR.

Closes #305.

## Not in this PR

The **backward step** (`.hzr_stepwise_backward_step`, `R/stepwise-step.R` ~704) still looks coefficients up by name, the root cause it shares with #304. It is recorded there:
- **Verified:** backward `hzr_stepwise()` on `~ fla + flag` stops with "Variable 'fla' expands to multiple coefficients".
- **Seen in this session, but not yet independently reproduced:** dropping `flag` from `~ fla + flag` tests `beta1`, the dummy. Whoever fixes #304 should reproduce it first.

The drop path has no refit to compare columns against, so it needs a different mechanism from this PR's.

**Expect conflicts with #302** (a draft stacked on #296). It edits `R/stepwise-step.R` at 227–240 and 401–431, right next to this PR's two call-site changes at the Wald candidate loop and the Wald fallback, and both PRs add `NEWS.md` bullets under 1.2.11. The two changes are independent: #302 changes the refit-failure *warnings*, and this PR changes *which coefficient is tested*. Whichever lands second needs a rebase, not a redesign. #302's log also queued this defect as an open item, and this PR addresses it.

**Keeping** the `.gitignore` commit (`59cfe06`), which adds `reports/`. The maintainer's global instructions put local findings reports in `reports/`, and those reports may quote PHI-adjacent content, so they must never be committed. Until now the directory was not ignored in this repo. It's a separate commit, so it can be dropped without touching the fix.

**`reports/` is now in `.Rbuildignore` too** (a follow-up to Copilot's review of `6fe307a`). `.gitignore` alone was not enough. `R CMD build` and `devtools::build()`/`check()` read `.Rbuildignore`, not `.gitignore`, so a tarball built from a working tree that contains reports packed them in. I reproduced this: `TemporalHazard/reports/` and its HTML file were in a tarball built from this branch's tree. With `^reports$` added, they are not, and `DESCRIPTION` is still there as a known positive. Release builds from `git archive` were never affected, because untracked files never reach the archive; only working-tree builds were. Drop this commit together with the `.gitignore` one if you'd rather not have either.


ehrlinger:
Re Copilot's review of `6fe307a` ("Exclude `reports/` from `.Rbuildignore`"): **confirmed and fixed in `0cea7f7`.**

This commit introduced the gap. `59cfe06` added `reports/` to `.gitignore` only. But `R CMD build`, `devtools::build()` and `devtools::check()` read `.Rbuildignore`, not `.gitignore`, so a build from a working tree that holds local findings reports would pack them into the tarball. Those reports may quote PHI-adjacent content.

**Reproduced before the fix.** `.Rbuildignore` had no `reports` entry. `R CMD build --no-build-vignettes` from this branch's tree produced a tarball containing:

```
TemporalHazard/reports/
TemporalHazard/reports/2026-09-13-stepwise-candidate-coef-name.html
```

**After adding `^reports$`,** the same build from the same tree has 0 `reports/` entries. `TemporalHazard/DESCRIPTION` is still present (the known positive), and `reports/` is still on disk in the tree, so the exclusion is doing the work.

Scope: release builds made from `git archive`, as AGENTS.md prescribes, were never affected, because untracked files never reach the archive. Only working-tree builds were. Lint is 0. No R code or tests changed.

Issue #300, opened 2026-09-14
predict(newdata) on a covariate-free fit mishandles an unused newdata column: errors, and a silent wrong answer for exponential lp/hazard


ehrlinger:
`predict(newdata = )` on a fit with **no covariates** (`Surv(...) ~ 1`) errors when `newdata` carries a column the model doesn't use. It should ignore that column, as fits with covariates already do. The failure is a loud error, not a wrong answer.

## Reproducer (observed on `main` at `a2ff8da`, 2026-09-13; **not yet re-verified on current `main`**)

```r
devtools::load_all()
set.seed(1)
d <- data.frame(time = rexp(200, 0.1), status = rbinom(200, 1, 0.7), age = rnorm(200, 60, 10))
f <- hazard(survival::Surv(time, status) ~ 1, data = d, dist = "weibull",
            theta = c(mu = 0.1, nu = 1), fit = TRUE)
predict(f, newdata = data.frame(time = 5), type = "survival")            # 0.7482958
predict(f, newdata = data.frame(time = 5, age = 70), type = "survival")  # ERROR: Number of parameters insufficient for predictor columns.
```

The expected result for the second call is 0.7482958, with `age` ignored.

## Status

PR #292 reworks the `newdata` design rebuild: `newdata` now supplies only the fit's data columns. That may change or fix this case. The owning session will re-run the reproducer on `main` after #292 lands and before changing anything, and will update this issue with the result.

ehrlinger:
**Re-verified on `main` at `9ec83e7` (after #292): not fixed, and one case is a silent wrong answer, not only an error.**

An exponential `Surv(time, status) ~ 1` fit (`theta = c(mu = 0.1)`), with `newdata = data.frame(time = 5, age = 70)`:

| type | `newdata` time only | with an unused `age = 70` column |
|---|---|---|
| `linear_predictor` | 0 | **-187.154**, exactly 70 × the fitted log rate (-2.67363) |
| `hazard` | 1 | **5.247e-82** |
| `survival` / `cumulative_hazard` | 0.708216 / 0.345005 | error ("Number of parameters insufficient for predictor columns.") |

[...]

PR #318, opened 2026-09-14
fix(predict): single-distribution predictions no longer carry a theta name (#309)

### LINKED ISSUE #309: Lognormal predict() survival/cumulative_hazard values are named after a theta element ("mu")
Lognormal `predict()` carries a coefficient's name onto its values: for a lognormal fit, `type = "survival"` and `type = "cumulative_hazard"` return values named `"mu"` (one per row). The values themselves are correct. This is the single-distribution version of #282 (multiphase predictions named after a theta element), which PR #289 fixes for multiphase fits only.

## Reproducer (confirmed on `main` at `9ec83e7`, 2026-09-14)

```r
devtools::load_all()
set.seed(1); d <- data.frame(time = rexp(200, 0.1), status = rbinom(200, 1, 0.7))
f <- hazard(survival::Surv(time, status) ~ 1, data = d, dist = "lognormal",
            theta = c(mu = 2, sigma = 1), fit = TRUE)
names(predict(f, newdata = data.frame(time = c(2, 5)), type = "survival"))
# "mu" "mu"
```

| family | survival | cumulative_hazard | hazard | linear_predictor |
|---|---|---|---|---|
| weibull, exponential, loglogistic | unnamed | unnamed | unnamed | unnamed |
| **lognormal** | **`"mu","mu"`** | **`"mu","mu"`** | unnamed | unnamed |

The source is `rep(th[1], ...)` at `R/hazard_api.R` around line 1740: a length-1 named element keeps its name when repeated. `unname()` it, as #289 does on the multiphase path.

A name on the result is cosmetic, but it propagates into data frames and anything that matches by name. Test: every family × every type returns unnamed values, for a covariate-free and a covariate fit, 1-row and multi-row `newdata`.

Found by the session fixing #300 (PR #308).

ehrlinger:
Closes #309

## What was wrong

For a single-distribution fit, `predict()` returned values named after a `theta` element. The values were right; the names were meaningless and followed the result into anything built from it.

The issue's table showed lognormal only. Probing `main` (`e4059a5`) over 4 families x {covariate-free, covariate} x {1, 3 rows} x 4 types x {`se.fit` FALSE, TRUE} = 128 cases found **34 named**:

| family | survival / cumulative_hazard, 3 rows | survival / cumulative_hazard, 1 row | `se.fit = TRUE`, 1 row | hazard / linear_predictor |
|---|---|---|---|---|
| lognormal | `"mu"` each value (no covariates) | `"mu"` | row name `"mu"` | unnamed |
| weibull, exponential, loglogistic | unnamed | `"mu"` | row name `"mu"` | unnamed |

## Census: where a theta name reaches the output

All in the `cumhaz_of` closures of `predict.hazard()`, `R/hazard_api.R` (the only place single-distribution predict does arithmetic on named `theta` elements; the eta-based `hazard`/`linear_predictor` branch goes through `as.numeric(x %*% beta)` and is unnamed):

1. **lognormal**: `rep(th[1], length(time))` names every element, for any n (the issue's line).
2. **lognormal**: `th[1] + as.numeric(x %*% beta_cand)` (covariate fit), and `/ exp(th[2])`, name a 1-row result.
3. **weibull**: `(th[1] * time) ^ th[2]` names a 1-row result (`th[2]` would too if `th[1]` did not).
4. **exponential**: `exp(th[1]) * time` names a 1-row result.
5. **loglogistic**: `exp(th[1]) * (time ^ exp(th[2]))` names a 1-row result.
6. **`se.fit = TRUE`**: `.hzr_predict_with_se()` builds `data.frame(fit = diff_fn(theta), ...)`, so a named 1-row `fit` became the row name. Fixed by 1-5, since `diff_fn` is `cumhaz_of`.

Because R takes a result's names from whichever operand has them, fixing operands one at a time leaves the next one leaking. So the fix `unname()`s each closure's **result** (four lines), as #289 does on the multiphase path (`R/likelihood-multiphase.R`).

## Evidence

- **Values unchanged:** after the fix, all 128 probe cases are **bit-identical** (`identical()`) to `main`, and **0 are named** (was 34).
- **New test** `tests/testthat/test-predict-single-dist-names.R`: every family x type x design x {1, 3} rows, point and `se.fit`, asserts `names()` is NULL, `se.fit` row names are `"1".."n"`, and values equal literals produced by `predict()` on `main` before the fix (tolerance 1e-12; the literals were checked `identical()` to the generator's output). Uses `fit = FALSE` models so the literals do not depend on each platform's optimizer; a grid-coverage count (64) guards against a silently empty loop. A second test runs the issue's fitted reproducer.
- **Mutation check:** removing `unname()` from the lognormal closure fails the test (18 failures); from the Weibull closure, 12 failures.

## Gates (at `cd5e83d`)

- `devtools::document()`: no changes to `man/` or `NAMESPACE`.
- `lintr::lint_package()`: **0 lints**.
- `spelling::spell_check_package(use_wordlist = TRUE)`: **0 words**.
- `NOT_CRAN=true devtools::test()` (`/Volumes/qhsstudies` mounted; `sum(passed)`, not `nb`): **5924 passed, 0 failed, 0 errors, 655 warnings, 6 skips**. The new file contributes 321 + 5 passes and 0 warnings. The 6 skips are exactly the known ones:
  - `test-bhdead-sas-parity.R` x3: bh.dead SAS fixture not available
  - `test-parity-c-binary.R` x1: legacy binary returned non-zero status 126
  - `test-weights-sas-parity.R` x2: weighted parity fixture `weighted-avc-fractional.rds` not found
- `R CMD check --as-cran` **with the manual**, from `git archive HEAD` extracted into a fresh `mktemp -d`: **Status: OK** (0 errors, 0 warnings, 0 notes; PDF and HTML manual OK). Tarball 3,283,235 bytes (`stat`); no `CLAUDE`/`AGENTS`/`reports/` entries.

Timings are not stamped: a sibling session's suite (worktree `vigilant-roentgen-0a4ab8`) ran concurrently with both the suite and the check.

## NEWS

Bug fixes bullet under 1.2.11 (no version bump). Also corrects #289's sentence "Single-distribution `predict()` already returned unnamed vectors", which was not true for lognormal, or for any family with one row.

## Seen in passing, not fixed here

A covariate-free fit refuses `newdata` carrying an unused column ("Number of parameters insufficient for predictor columns"). That is #300 (PR #308).


ehrlinger:
**Copilot review triage (review 5219979254, made on `cd5e83d`, before the merge of `main`).** Verdict "approval recommended", 0 inline comments, one suppressed comment. Recording it here because a suppressed comment has no thread.

**Finding:** the test header (`tests/testthat/test-predict-single-dist-names.R`, line 2) says "The shape parameters are named elements of theta". Exponential's `theta[1]` is a log rate, not a shape.

**Verdict: correct, and it goes further than flagged.** The name that leaked is always `"mu"`, which is `theta[1]`, and that is never a shape parameter. Reproduced by running `predict(type = "cumulative_hazard")` at `e2cc377`, `newdata` time 2, `fit = FALSE`:

| family | theta | H(2) | closed form | so `theta[1]` is |
|---|---|---|---|---|
| exponential | `mu = log(0.2)` | 0.4 | `exp(mu) * t` | the log rate |
| weibull | `(mu, nu) = (0.2, 1)`, then `nu = 2` | 0.4, then 0.16 | `(mu * t)^nu` | a scale (`nu` is the shape) |
| lognormal | `(mu, sigma) = (1, 0)` | 0.4771937 | `-log(pnorm(-(log(2) - 1)))` | the location |

The wording came from the internal `.hzr_shape_parameter_count()`, which counts every non-covariate parameter as "shape". As prose it is wrong in the test header and also in the code comment at `R/hazard_api.R:1713`, which Copilot did not flag. The NEWS bullet ("`mu` is a named element of `theta`") is accurate.

**Severity:** comment-only. No behaviour change and no wrong answer.

**Disposition:** not changed on this PR, so the verified green head `e2cc377` stays as it is. Both comments are fixed in a follow-up via #314, which edits `R/hazard_api.R` anyway.

